PR #3, opened 2025-01-20
Add working draft sections of §6.10 and full 2018 and 2024 C standards

gklimowicz:
It's helpful to have the full standards and the subsection on the preprocessor. These come from the freely-available working drafts.


gklimowicz:
@bonachea I don't do much "collaborative" work these days, but I would like to get into the practice of having people review changes I propose to this repo. Please advise if you have any issues.

gklimowicz (commented):
".SD_Store" in commit comment should be ".DS_Store".

gklimowicz:
@certik @bonachea FYI: I am closing this pull request to resubmit with a corrected comment and to only do one thing at a time. I think that's better style.

bonachea:
@gklimowicz I realize you already closed this PR, but I would strongly advise against uploading (sections of) IEEE copyrighted publications to this repo (or any public location).

gklimowicz:
@bonachea Agreed. I usually make sure I only put up copies of working drafts, which are publicly available. I've been told those are OK to post (like we do with the Fortran drafts). There may be something in the deleted PR that slipped through, though.

It may be more cautious to just put up the §6.10 text. I consider that "fair use", but I am not a lawyer. Let me know what you think.

bonachea:
@gklimowicz :
I agree that working drafts should be ok, and by extension 6.10 sections of those. 

Let's avoid posting any portion of an official standards document carrying the official IEEE copyright header. It might be fair use but I'd rather not risk it.

This closed PR still contains a copy of PDFs breaking that rule (despite the branch having been deleted) so I plan to also delete this PR entirely, unless you object.

gklimowicz:
@bonachea Delete away. (I am on my phone at the moment.)

Issue #14, opened 2025-01-31
Types in preprocessor expressions: is everything an integer?


gklimowicz:
In the C23 standard, a preprocessor _constant-expression_ is evaluated to an integer, with zero meaning _false_, and non-zero meaning _true_. This includes the comparison operators (`==  != < <= >  >=`) and logical _and_ (`&&`), _or_ (`||`), and _not_ (`!`) operators.

In Fortran, the analogous comparison and logical operators evaluate to `LOGICAL` values: `.TRUE.` or `.FALSE.`.

To maintain consistency with C semantics, the preprocessor should probably evaluate comparison operators (`.EQ.  .NE.  .LT.  .LE.  .GT.  .GE.  .EQV.  .NEQV.`) as if they return integer zero for _false_ and `1` (or an unspecified non-zero value) for _true_.

Similarly, the logical operators (`.AND.  .OR.  .NOT.`) should probably also evaluate to integer values representing _false_ and _true_.

The upside of this is faithfulness to the C standards. 

A potential downside is that it may be difficult to add back any kind of `LOGICAL` typing of operators in some future version of the standard.

bonachea:
I'm strongly in favor of this. 

IMO the most compelling argument for treating everything as an INTEGER is that the "alternative" of maintaining disparate INTEGER and LOGICAL types in preprocessor conditional compilation expressions probably makes it far more awkward (impossible?) to combine LOGICAL and INTEGER values in a single preprocessor conditional compilation expression. This is _because_ none of the Fortran intrinsics are available in preprocessor conditional compilation expressions to convert between types. Thus it's best to treat everything as a single (maximal integer) type, which preserves expressiveness without introducing pointless type conflicts.

jeffhammond:
Is this the Fortran preprocess or the C preprocess in Fortran?  You have to decide one way or the other, unambiguously, or it's going to be a disaster.

This issue cannot be evaluated separately from https://github.com/j3-fortran/fpp-202y/issues/13.  They are the same issue, because types and operators are dependent on one another.

bonachea:
@jeffhammond said:
> Is this the Fortran preprocess or the C preprocess in Fortran? You have to decide one way or the other, unambiguously, or it's going to be a disaster.
> 
> This issue cannot be evaluated separately from [#13](https://github.com/j3-fortran/fpp-202y/issues/13). They are the same issue, because types and operators are dependent on one another.

@jeffhammond What exact technical change are you advocating here?

[...]

PR #15, opened 2025-02-01
Add missing Fortran operators to preprocessor #if, #elif expressions

### LINKED ISSUE #13: Expand operators in preprocessor expressions to include Fortran operators
When evaluating expressions in `#if` and `#elif`, paper 24-177r1 list the following operators from the C23 standard:
```
&   &&   
|   ||  
==  !=
<   <=
>   >=
+   -
/   *   
!   
^    ~  
( )
```

It also included the following Fortran operators:
```
/= 
.AND. 
.OR.
.NOT.
```
To allow macro definitions that can be used in preprocessor expression evaluation and also in token replacement in Fortran statement text, we should also make the following Fortran operators available in expression evaluation:
```
.EQ.    .NE.  
.LT.    .LE.
.GT.    .GE.
.EQV.   .NEQV.
```

gklimowicz:
This allows the same #define-d value to de used in Fortran source and in directive expressions.

Fixes #13.

### DIFF
--- a/drafts/25-revise-24-177 Fortran preprocessor requirements.txt
+++ b/drafts/25-revise-24-177 Fortran preprocessor requirements.txt
@@ -122,7 +122,8 @@ preprocessor will include the text from Fortran INCLUDE lines
 ---------------------------------------------------
     - The "defined" operator
     - From C: && || == != < > <= >= + / * ! & | ^ ~ ( )
-    - From Fortran: = /= .AND. .OR. .NOT.
+    - From Fortran: = /= .AND. .OR. .NOT. .EQ. .NE.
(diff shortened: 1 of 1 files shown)


bonachea (commented):
As already expressed in #16, I'd like us to address that PR and issue #14 first, and once that's merged consider this orthogonal expansion of operators.

bonachea (changes requested):
As already expressed in #16, I'd like us to address that PR and issue #14 first, and once that's merged consider this orthogonal expansion of operators.

gklimowicz:
After discussion with JoR today before meeting #235, we agreed to remove references to all Fortran operators and constants from the 25-114 requirements.

Should a set of compelling use cases emerge, we can revisit this.

PR #16, opened 2025-02-01
Add more rules on preprocessor expressions and their types

gklimowicz:
Change all preprocessor expressions for #if, #elif to evaluate to integer values. Define the relationship of Fortran logical constants and operators to their C counterparts.

Addresses #14.

### DIFF
--- a/drafts/25-revise-24-177 Fortran preprocessor requirements.txt
+++ b/drafts/25-revise-24-177 Fortran preprocessor requirements.txt
@@ -11,6 +11,8 @@ References: 95-257 Conditional Compilation: The FCC Approach.txt
             24-109 On Fortran awareness in a Fortran preprocessor.txt
             24-177r1 Fortran preprocessor requirements.txt
             Tutorials/Preprocessor Take 2.pptx
+            ISO/IEC 9899:2023 Programming languages — C
+                   working draft N3096
 
 
 1. Introduction
@@ -43,30 +45,6 @@ We plan to take a similar approach for defining FPP. This should
 simplify the explanation of the expected behavior of any given
 implementation.
 
-The directive language accepted by FPP is based on the syntax of CPP. It
-is not Fortran, but can contain arbitrary Fortran tokens. It differs
-from C and Fortran syntax in the following ways:
-    1. FPP's directives and token recognition are case sensitive.
-    2. FPP treats blanks adjacent to tokens as significant, even in
-       fixed-form source files.
-    3. FPP directive lines accept backslash (\) for line continuations.
-    4. FPP does not recognize fixed-form (column 6) or free-form (&)
-       continuations on directive lines.
-    5. FPP does not recognize '!' as initiating a comment on directive
-       lines.
-    6. FPP does not recognize '//' as initiating a comment on directive
-       lines.
-    7. FPP recognizes /* ... */ C-style comments on directive lines.
-    8. Expressions in #if and #elif directives contain constructs from
-       both Fortran and CPP.
-    9. Expressions in #if and #elif directives that evaluate to LOGICAL
-       or INTEGER values. .TRUE. or a non-zero integer are considered
-       true.
-    10. Undefined identifiers are treated as zero, as in CPP.
-    11. Integer expressions in preprocessor directives are calculated in
-       the maximum precision the processor supports. (There are no KIND
-       specifiers on integer constants in the preprocessor.)
-
 
 3. FPP Phase 1: Line conjoining
 -------------------------------
@@ -87,14 +65,61 @@ The directive processing phase is analogous to CPP phase 4. Preprocessor
 directives are executed. Macros are expanded in non-preprocessor lines
 (Fortran source lines).
 
-Macros are expanded in comment lines that appear to be pragma directives
-(e.g. '!$omp' or '!$acc' or additional processor-specific choices).
+The directive language accepted by FPP is based on the syntax of CPP. It
+has syntax that differs from Fortran, but macros can expand to include
+arbitrary Fortran tokens. It differs from C and Fortran syntax in the
+following ways.
 
-#pragma lines are expanded into Fortran 'pragma' lines (a new feature).
+Token recognition:
+    1. FPP's directives and token recognition are case sensitive.
+    2. FPP treats blanks adjacent to tokens as significant, even in
+       fixed-form source files.
+
+Line continuation in directives:
+    1. FPP directive lines accept backslash (\) for line continuations.
+    2. FPP does not recognize fixed-form (column 6) or free-form (&)
+       continuations on directive lines.
+
+Comment handling:
+    1. FPP does not recognize '!' as initiating a comment on directive
+       lines. In #if and #elif directive expressions, '!' is interpreted
+       as the C 'not' operator.
+    2. FPP does not recognize '//' as initiating a comment on directive
+       lines. '//' can be used to construct macro definitions that
+       contain Fortran string concatenation.
+    3. FPP recognizes /* ... */ C-style comments on directive lines.
+       /* ... */ comments are not recognized in (non-directive)
+       Fortran source lines.
+    4. Macros are expanded in comment lines that appear to be directives
+       (processor-specific, such as '!$omp', or '!$acc', or others).
+
+Constant expressions in #if and #elif:
+    1. Expressions in #if and #elif directives allow operators from
+       both Fortran and CPP.
+    2. Expressions in #if and #elif directives must be integer constant
+       expressions as specified for CPP (with the extensions described
+       below), and evaluate to INTEGER values. As in CPP, zero values
+       are treated as 'false'. Non-zero values are treated as 'true'.
+    3. .FALSE. and .false. are treated as the integer 0. .TRUE. and .true.
+       are treated as the integer 1.
+    4. Any undefined identifiers that remain after macro expansion
+       (including those lexically identical to keywords or intrinsics)
+       are treated as zero, as in CPP.
+    5. C character constants (such as 'A', '\n') are treated as
+       integer values, as they are in CPP.
+    6. The Fortran operators .AND., .OR., ,NOT., =, and /= evaluate
+       to the same values as the C operators &&, ||, !, ==, and !=,
+       respectively.
+    7. There are no KIND specifiers on integer constants in the
+       preprocessor.
+    8. Integer expressions in preprocessor directives are evaluated using
+       the maximum precision the processor supports.
 
 
 4.1. Directives accepted by the preprocessor
 --------------------------------------------
+The following preprocessor directives will have the same semantics
+as defined in the C23 edition of the C programming language standard.
     #line
     #define  including function-like macros and those
              that uses the ellipsis notation in the parameters
@@ -107,7 +132,9 @@ Macros are expanded in comment lines that appear to be pragma directives
     #pragma
 
 Just as #include lines interpolate the source from other files, the
-preprocessor will include the text from Fortran INCLUDE lines
+preprocessor will include the text from Fortran INCLUDE lines.
+Text interpolated by INCLUDE lines will be treated as if it had
+been included via #include.
 
 
 4.2 Tokens accepted on #define directives


drafts/25-revise-24-177 Fortran preprocessor requirements.txt
```diff
@@ -58,13 +58,22 @@ from C and Fortran syntax in the following ways:
...
-    10. Undefined identifiers are treated as zero, as in CPP.
-    11. Integer expressions in preprocessor directives are calculated in
-       the maximum precision the processor supports. (There are no KIND
-       specifiers on integer constants in the preprocessor.)
+    9. Expressions in #if and #elif directives evaluate to INTEGER values.
+       Zero values are treated as 'false'. Non-zero values are treated as
+       'true'.
+    10 .FALSE. is treated as the integer 0. .TRUE. is treated as the
+       integer 1.
+    11. Undefined identifiers are treated as zero, as in CPP.
```
bonachea, line 66: Minor clarification, inspired by C spec wording:
```suggestion
    11. Any undefined identifiers that remain after macro expansion (including those 
        lexically identical to keywords or intrinsics) are treated as zero, as in CPP.
```

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
```diff
@@ -58,13 +58,22 @@ from C and Fortran syntax in the following ways:
...
+    12. C character constants (such as 'A', '\n') are treated as
+        integer values, as they are in C.
+    13. The Fortran operators .AND., .OR., ,NOT., .EQ., .NE., .EQV., and
+        .NEQV. calculate the save values as the C operators &&, ||, !, ==,
+        !=, ==, and !=, respectively.
+    14. The Fortran operators .LT., .LE., .GT., and .GE. calculate the
+        same values as the C operators <, <=, >, and >=, respectively.
+    15. Integer expressions in preprocessor directives are calculated in
+        the maximum precision the processor supports. (There are no KIND
+        specifiers on integer constants in the preprocessor.)
```
bonachea, line 76: This feels like it should be two separate rules. IIUC the parenthetical is actually prohibiting certain syntax, so it should not be parenthetical.

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
```diff
@@ -58,13 +58,22 @@ from C and Fortran syntax in the following ways:
...
+       Zero values are treated as 'false'. Non-zero values are treated as
+       'true'.
+    10 .FALSE. is treated as the integer 0. .TRUE. is treated as the
+       integer 1.
+    11. Undefined identifiers are treated as zero, as in CPP.
+    12. C character constants (such as 'A', '\n') are treated as
+        integer values, as they are in C.
+    13. The Fortran operators .AND., .OR., ,NOT., .EQ., .NE., .EQV., and
+        .NEQV. calculate the save values as the C operators &&, ||, !, ==,
+        !=, ==, and !=, respectively.
```
bonachea, line 71: Here you're assuming we also accept the suggestion in issue #13 to add these additional Fortran operators to the draft. If we were to merge this PR alone as-is, these new rules would be inconsistent with the earlier list of operators.

This PR purports to address only #14 (which IMO has a strong stand-alone motivation), and I'd like to keep it focused on that proposal without conflating the orthogonal issue #13 (where the motivation may be seen as potentially weaker).

Please decouple these two PRs. I suggest removing the "new" operators above and getting this PR into a mergeable state to address solely #14, and then PR #15 can later introduce the obvious extension for the operators proposed by #13.

```suggestion
    13. The Fortran operators .AND., .OR., ,NOT., =, /= evaluate to the same values as 
        the C operators &&, ||, !, ==, and !=, respectively.
```

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
```diff
@@ -58,13 +58,22 @@ from C and Fortran syntax in the following ways:
...
+    14. The Fortran operators .LT., .LE., .GT., and .GE. calculate the
+        same values as the C operators <, <=, >, and >=, respectively.
```
bonachea, line 73: Same comment here, please move this new rule to PR #15

```suggestion
```

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
```diff
@@ -58,13 +58,22 @@ from C and Fortran syntax in the following ways:
...
-       the maximum precision the processor supports. (There are no KIND
-       specifiers on integer constants in the preprocessor.)
+    9. Expressions in #if and #elif directives evaluate to INTEGER values.
+       Zero values are treated as 'false'. Non-zero values are treated as
+       'true'.
+    10 .FALSE. is treated as the integer 0. .TRUE. is treated as the
+       integer 1.
+    11. Undefined identifiers are treated as zero, as in CPP.
+    12. C character constants (such as 'A', '\n') are treated as
+        integer values, as they are in C.
```
bonachea, line 68: This run-on list of rules needs to be restructured to clarify that the rule above (and possibly others?) only apply to the conditional expression in `#if/#elif`. For example this does not impact macro expansion in source.

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
```diff
@@ -58,13 +58,22 @@ from C and Fortran syntax in the following ways:
...
-    9. Expressions in #if and #elif directives that evaluate to LOGICAL
-       or INTEGER values. .TRUE. or a non-zero integer are considered
-       true.
-    10. Undefined identifiers are treated as zero, as in CPP.
-    11. Integer expressions in preprocessor directives are calculated in
-       the maximum precision the processor supports. (There are no KIND
-       specifiers on integer constants in the preprocessor.)
+    9. Expressions in #if and #elif directives evaluate to INTEGER values.
+       Zero values are treated as 'false'. Non-zero values are treated as
+       'true'.
```
bonachea, line 63: When it comes time to actually write spec wording for this, it's not enough to say the expression "evaluate to INTEGER values". There must be additional constraints to rule out (for example) character strings in preprocessor conditional expressions.

For now it might be sufficient to say something like 
```suggestion
    9. Expressions in #if and #elif directives must be integer constant expressions as in CPP
       (with the extensions described below), and evaluate to INTEGER values.
       As in CPP, zero values are treated as 'false'. Non-zero values are treated as
       'true'.
```

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
```diff
@@ -58,13 +58,22 @@ from C and Fortran syntax in the following ways:
...
+    10 .FALSE. is treated as the integer 0. .TRUE. is treated as the
+       integer 1.
```
bonachea, line 65: What about `.false.` and `.true.`?

bonachea (changes requested):
Thanks for making a great start on this!

This is a non-trivial change and I have a correspondingly large amount of feedback. I may have even more feedback after further updates and further thought.

However I think we're on the right track here, and as I've already said I really want to see us adopt the core changes proposed in #14.

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
gklimowicz: I will add `.false.` and `.true.` explicitly.

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
gklimowicz: In fact, all this feels like it should move down to (current) section 4, Directive Processing.

And since I still plan to write the mythical paper on compiler phases, section 2 (The basic idea: phases before the "processor") should be deleted. Section 3, too. (FPP Phase 1: Line conjoining).

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
gklimowicz: Will incorporate this change.

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
gklimowicz: Will incorporate this change.

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
gklimowicz: Will incorporate this change.

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
gklimowicz: Will create two rules.

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
```diff
@@ -87,14 +65,59 @@ The directive processing phase is analogous to CPP phase 4. Preprocessor
...
+       /* ... */ comments are not recognized in (non-directive)
+       Fortran source lines.
+    4. Macros are expanded in comment lines that appear to be directives
+       (processor-specific, such as '!$omp', or '!$acc', or others).
+
+Constant expressions in #if and #elif:
+    1. Expressions in #if and #elif directives allow operators from
+       both Fortran and CPP.
+    2. Expressions in #if and #elif directives must be integer constant
+       expressions as in CPP (with the extensions described below), and
```
aury6623, line 98: In Fortran standardese, an "integer constant expression" is any constant expression that has an overall/result type of integer. For example, `FLOOR(SUM([1.0,2.0,3.0])) + LEN("ABC"//"DEF")` is a (scalar!) integer constant expression with the value 12, just going by the existing wording in the standard.

Obviously this sentence I've highlighted isn't going to go in the standard verbatim anyway, but my point is just that when it comes time for edits, we might need to be careful about reusing words like "constant expression". In this case we'd at least want to say something like "after macro expansion and substitution of .FALSE., .TRUE., and char constants, all primaries in the expression shall be integer literals".

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
```diff
@@ -87,14 +65,59 @@ The directive processing phase is analogous to CPP phase 4. Preprocessor
...
+       evaluate to INTEGER values. As in CPP, zero values are treated as
+       'false'. Non-zero values are treated as 'true'.
+    3. .FALSE. and .false. are treated as the integer 0. .TRUE. and .true.
+       are treated as the integer 1.
+    4. Any undefined identifiers that remain after macro expansion
+       (including those lexically identical to keywords or intrinsics)
+       are treated as zero, as in CPP.
+    5. C character constants (such as 'A', '\n') are treated as
+       integer values, as they are in CPP.
```
aury6623, line 107: Like my previous comment, we may need to avoid saying "character constant" when we mean something like `'A'` because in Fortran, a character constant is `"ABCDE"`. Perhaps we'll have to call them "char constant"? \*shrug\*

bonachea (changes requested):
Thanks for all the fixes!

This is definitely a big improvement, and I just have a few relatively minor outstanding suggestions.

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
bonachea: I agree with the spirit of @aury6623 's observation, but not the proposed resolution.

In this wording: (emphasis added)

> must be integer constant expressions **as in CPP**

"as in CPP" was intended to convey "as specified for CPP", meaning both the C language definition of the term ["integer constant expression"](https://en.cppreference.com/w/c/language/constant_expression#:~:text=(since%20C99)-,Integer%20constant%20expression,-An%20integer%20constant), and the [additional constraints that CPP places on #if and #elif directive expressions](https://en.cppreference.com/w/cpp/preprocessor/conditional#:~:text=Condition%20evaluation).

@aury6623 would this minimal change resolve your concern for the purposes of the requirements document?

```suggestion
    2. Expressions in #if and #elif directives must be integer constant
       expressions as specified for CPP (with the extensions described below), and
```

Presumably when it comes time to write actual edits we won't be writing as a diff against CPP (done here for brevity), so such terminology differences should not even arise.


> say something like "after macro expansion and substitution of .FALSE., .TRUE., and char constants, all primaries in the expression shall be integer literals".

This proposed wording would be incorrect for several reasons: the `defined` operator, and also rule 4 below.

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
```diff
@@ -87,14 +65,59 @@ The directive processing phase is analogous to CPP phase 4. Preprocessor
...
+
+Line continuation in directives:
+    1. FPP directive lines accept backslash (\) for line continuations.
+    2. FPP does not recognize fixed-form (column 6) or free-form (&)
+       continuations on directive lines.
 
-#pragma lines are expanded into Fortran 'pragma' lines (a new feature).
+Comment handling:
+    1. FPP does not recognize '!' as initiating a comment on directive
+       lines. '!' is interpreted as the C 'not' operator.
```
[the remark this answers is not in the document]
aury6623: I actually just tested this.
- Line 1: not operator, as intended

- Line 2: **Currently, different predecessors handle this differently, in a pretty significant way that I think we'll need to address.** For example:
```F90
#define PI 3.141 ! close enough for government work
print *, PI, "foo"
end
```
ifx and nagfor compile and execute this program and print "3.141000     foo" (whitespace differs due to print semantics, but same contents)
flang ("flang-trunk" on godbolt.org) compiles and executes and  prints "3.141".
gfortran compiles and executes and prints "3.14100003". Ignoring the floating point precision difference, gfortran and flang don't print "foo" because it performs the macro substitution like this:
```
print *, 3.141 ! close enough for government work, "foo"
````
(Output of `-E` flag with gfortran. `-E` with flang just outputs "print *, 3.141" so I guess their preprocessor also deletes comments.)
They treat the "!" just as any other text, and so it substitutes the "! close enough for government work" into the middle of the statement.
Both the gfortran/flang and nagfor/ifx approaches have merits, but IMO the expected behavior of a Fortran programmer would be the nagfor/ifx behavior. It seems extremely easy to make unnoticed mistakes if "! comment" gets substituted literally.
So, I prefer we say that "!" is treated as a comment by the preprocessor on `#define` lines.

- Line 3: gfortran, ifx, nagfor, and flang all seem to just ignore all text after an `#endif`. You can even write
```
#endif foo
```
(no comment indicator) and none of them complain. It doesn't matter that much what we pick, but my thought on the matter is that the standard should say something equivalent to "text on the life after `#endif` is treated as a comment".

----

Also what about this? 😄 
```
#include "file.f90" hello
```
ifx and nagfor both silently ignore text after the file name string on the `#include` line. gfortran ignores it, but with a warning: "extra tokens at end of #include directive". I'm not sure about flang on godbolt.org because I don't see a way to  have multiple files, but based on the error message it gave, I think it behaves like ifx and nagfor:
```
/app/example.f90:1:10: error: #include: Source file 'file.f90' was not found
  #include "file.f90" hello
           ^^^^^^^^^^
```
So, I figure the standard can just say text after the file name on an `#include` is ignored, just like with `#endif`.

File: drafts/25-revise-24-177 Fortran preprocessor requirements.txt [hunk 11]
```diff
@@ -87,14 +65,59 @@ The directive processing phase is analogous to CPP phase 4. Preprocessor
 directives are executed. Macros are expanded in non-preprocessor lines
 (Fortran source lines).
 
-Macros are expanded in comment lines that appear to be pragma directives
-(e.g. '!$omp' or '!$acc' or additional processor-specific choices).
+The directive language accepted by FPP is based on the syntax of CPP. It
+is not Fortran, but can contain arbitrary Fortran tokens. It differs
```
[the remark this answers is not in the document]
aury6623: My thought in regard to "It is not Fortran" is that it helps the uninitiated reader (the people from the standard committee who will read this) understand where we're coming from. FPP lines are "not Fortran" in the sense that they do not follow any of the existing rules set in the Fortran standard, and so we can't approach it as if they did. The syntax of a FPP line is essentially "C with some Fortran extensions".

The "can contain arbitrary Fortran tokens" part seems unnecessary, though.

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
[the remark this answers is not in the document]
bonachea: @aury6623 said:
> different predecessors handle this differently,

Thanks for the report.

It sounds like this is yet another "dark corner" where current practice has significantly diverged in the vacuum created by a lack of standardization. IMHO resolving such discrepancies to improve the user experience is the most important motivation for our overall effort.

That being said, I don't think it's productive for the requirements document to attempt to dive into such a fine-grained level of detail. This one of a long list of discrepancies to be resolved when we are actually drafting edits.

In any case, the observation definitely seems orthogonal to the goal of this PR.

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
aury6623: Ah, I interpreted "as in CPP" to mean something like "CPP only allows integers in their preprocessor expressions, so we will too".

I don't know if "as specified for CPP" is even sufficient for someone who does not know that "integer constant expression" has a different definition in C. I think it would need to be more explicit, like:
```
    2. Expressions in #if and #elif directives must be integer constant
       expressions, using the definition of "integer constant expression" as defined by C, with the extensions described below, and
```
though of course that is quite verbose.

Another option to avoid using the term "integer constant expression" entirely:
```suggestion
    2. Expressions in #if and #elif directives must only contain integer primaries,
       with the extensions described below, and
```

---

> Presumably when it comes time to write actual edits we won't be writing as a diff against CPP

Definitely not. I think the closest we'd be able to get is mentioning the C preprocessor in a note.

> This proposed wording would be incorrect for several reasons: the defined operator, and also rule 4 below.

I didn't expect my wording to be perfect or ready for the standard, hence the wording prefixed with "at least want to say something like". However, I expect that our final edit will have to include the words "shall only contain integer primaries" with some sort of list of caveats to accommodate things like `defined` and `.TRUE.` and `'A'`. There's no way we'd be allowed to say anything close to "shall be an integer constant expression as defined by \<reference to C standard\>".

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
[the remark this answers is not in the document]
aury6623: Sure. I just think we might not want to say
```
    1. FPP does not recognize '!' as initiating a comment on directive
       lines. '!' is interpreted as the C 'not' operator.
```
here.

Perhaps only mention `#if` for now and sneak away from mentioning other cases until later.
```
   1. FPP does not recognize '!' as initiating a comment on #if lines.
       '!' is interpreted as the C 'not' operator.
```

Or be bold and add to that:
```
   1a. FPP recognizes '!' as initiating a comment on #define lines.
```

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
bonachea: @aury6623 suggested:
> #if and #elif directives must only contain integer primaries,

"integer constant expression" is clearly defined in the C spec and used in about 80 places throughout C23. We could cite the exact relevant section (C23 6.10.2-3) if we think that helps. For discussion purposes, here is the actual C23 prose:

> The expression that controls conditional inclusion shall be an integer constant expression except that: identifiers (including those lexically identical to keywords) are interpreted as described subsequently in this subclause197) and it may contain zero or more `defined` macro expressions, `has_include` expressions, `has_embed` expressions, and/or `has_c_attribute` expressions as unary operator expressions.

The compound phase "integer primaries" you suggest (or "integer primary") does not seem to appear anywhere in F23, which means that wording seems to leave the door open to more misunderstanding. Also unlike the C reference above, one is left to scratch their head about macro expansion of identifiers, treatment of undefined identifiers, the `defined` operator, and logical expressions, etc.

gklimowicz: Many of these details will be explicit in the syntax. I don't want to go overboard trying to establish too detailed a set of semantic rules if we don't have the syntax to hang it on.

For the purposes of requirements, "it's integers all the way down". And C23 is the rationale.

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
gklimowicz: I think this, too, will be explicit in the expression syntax. We will want a lexical analyzer that is smart enough to say that in the context of preprocessor expressions, 'A' is a character constant, "A" is a string, 'AA' is not a character constant. The first is legal in an expression, but the others are not.

In a `#define`, all are legal tokens that could be expanded from a macro name or macro function.

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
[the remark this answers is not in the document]
gklimowicz: Thanks for this. What I'm trying to say is based on the following:

- The preprocessor is not doing string manipulation. It's doing token manipulation.
- Directives don't look like Fortran; they look like what you would see in CPP. So, it's not Fortran and it's not CoCo, and ...
- In the replacement text of a macro name or function-like macro, any legal Fortran tokens are allowed. This is behavior that expands on what CPP allows.
- In the replacement text of a macro name or function-like macro, a subset of C tokens may also make sense. (For example, in an identifier or function that will be expanded in a `#if` or `#elif`.)
- As Aury points out, nothing has to be syntactically valid until completely expanded. And it's not the preprocessor that cares. (later syntactic analysis for proper Fortran source will.)

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
```diff
@@ -87,14 +65,59 @@ The directive processing phase is analogous to CPP phase 4. Preprocessor
...
+       'false'. Non-zero values are treated as 'true'.
+    3. .FALSE. and .false. are treated as the integer 0. .TRUE. and .true.
+       are treated as the integer 1.
+    4. Any undefined identifiers that remain after macro expansion
+       (including those lexically identical to keywords or intrinsics)
+       are treated as zero, as in CPP.
+    5. C character constants (such as 'A', '\n') are treated as
+       integer values, as they are in CPP.
+    6. The Fortran operators .AND., .OR., ,NOT., =, and /= evaluate
+       to the same values as the C operators &&, ||, !, ==, and !/,
```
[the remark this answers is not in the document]
gklimowicz: Whoops. I'll fix this. (I won't commit this here, as I tend to get all tangled up merging it from `origin/preprocessor-expressions-are-integer-expressions`.)

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
```diff
@@ -138,11 +163,11 @@ preprocessor will include the text from Fortran INCLUDE lines
...
-__STDF__ is an analog to __STDC__ in C and __cplusplus in C++.  Its primary
-role is to provide preprocessor-visible and vendor-independent identification
-of the underlying target language (i.e., "the processor is Fortran"), which
-enables one to write multi-language header files with conditional compilation
-based on language.
+__STDF__ is an analog to __STDC__ in C and __cplusplus in C++. Its
+primary role is to provide preprocessor-visible and vendor-independent
+identification of the underlying target language (i.e., "the processor
+is Fortran"), which enables one to write multi-language header files
+with conditional compilation based on language.
```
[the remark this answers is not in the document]
gklimowicz: In the future, I will work to suppress my OCD tendencies in this regard. It's a curse, especially when working with others.

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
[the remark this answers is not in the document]
gklimowicz: The handling of `!` is complicated, and I wanted to start with a simple rule first.

For example, I don't think the following should define `NOT_NASA` as empty:
```
#define NOT_NASA ! NASA
#if NOT_NASA
```

So, syntactically, I think the only allowable comment indicators FPP should accept on directives of any kind are `/* ... */` style. We can't allow `//` either, for Fortran token reasons.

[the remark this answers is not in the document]
gklimowicz: With respect to `#include`, I was following the C preprocessor here, where a `#include` is allowed to contain things like `"foo.h"` or `<foo.h>`, or a sequence of preprocessing tokens that are expanded to match one of these two forms. All of these are immediately followed by a new-line. See 6.10.2 paragraph 4 in the C23 working draft. (This behavior goes back to C99.)

If we want to be stricter than CPP, I would recommend the C90 statement that only sequences like `"foo.h"` and `<foo.h>` be allowed.

CPP does not appear to allow gratuitous text after the file name as in `#include "file.f90" hello`.

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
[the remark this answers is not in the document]
bonachea: I'm not sure I agree with everything in @gklimowicz 's reply above, but it's not important to resolve today.

I propose this concrete change to the wording to convey what I think we mean, without diving into too many details:

```suggestion
has syntax that differs from Fortran, but macros can expand to include arbitrary Fortran tokens. It differs
```

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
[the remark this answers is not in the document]
bonachea: IIUC this typo was introduced in this PR branch, so it should be fixed here prior to merge

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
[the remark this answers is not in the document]
bonachea: > I put the original text back.

Thanks, but I don't yet see a commit reverting the whitespace change?

drafts/25-revise-24-177 Fortran preprocessor requirements.txt
gklimowicz: @aury6623 what would you change in the context of "requirements" (not syntax at this time)? I'm comfortable leaving as it is for now.

aury6623: I'll accept that the current wording is probably fine for requirements. I don't think anyone will be confused.

bonachea (approved):
LGTM!
Thanks for the all the fixes.

aury6623 (approved):

gklimowicz:
Thanks for all the diligent reviews.

PR #22, opened 2025-02-04
Avoid introducing new predefined macros

aury6623:
We may wish to add these later, but I think we should avoid bringing these up now. I'm fairly certain that these will just become a distraction at the meeting.

I'm also of the opinion that we shouldn't add `__STDF__` at all. No FPP that I've tested supports it already, so including it isn't an attempt to standardize existing behavior; it's entirely a new feature. And it's a new feature that probably isn't very useful, as we discussed in last week's meeting.

`__VA_ARGS__` I think is neat and possibly worth including, but out of nagfor, ifx, gfortran, and flang, only flang supports it currently. So it's not like it's already a common feature.

If we're introducing completely (or nearly completely) new behavior, I think it makes more sense to give them dedicates papers to move at the committee meeting where we can debate only that one new feature. Otherwise, we'll end up debating it anyway, but when we should be discussing everything else in this paper!

(This argument also applies to other "new" features like proposing a new `#pragma`, though for whatever reason, proposing `#pragma` seems more acceptable to me. Probably my biases 😄 But I wouldn't mind seeing `#pragma` removed from this paper too and reintroduced in a separate feature paper where we can discuss it specifically.)

### DIFF
--- a/drafts/25-revise-24-177 Fortran preprocessor requirements.txt
+++ b/drafts/25-revise-24-177 Fortran preprocessor requirements.txt
@@ -132,17 +132,6 @@ preprocessor will include the text from Fortran INCLUDE lines
     __FILE__
     __DATE__
     __TIME__
-    __STDF__
-    __VA_ARGS__  in the replacement-list of a function-like macro
-      that uses the ellipsis notation in the parameters.
-    __VA_OPT__  in the replacement-list of a function-like macro
-      that uses the ellipsis notation in the parameters.
-
-__STDF__ is an analog to __STDC__ in C and __cplusplus in C++.  Its primary
-role is to provide preprocessor-visible and vendor-independent identification
-of the underlying target language (i.e., "the processor is Fortran"), which
-enables one to write multi-language header files with conditional compilation
-based on language.
 
 4.5. Fortran awareness during macro expansion
 ---------------------------------------------


kc9jud:
> And it's a new feature that probably isn't very useful, as we discussed in last week's meeting.

I seem to remember deciding that __STDF_VERSION__ isn't useful, but that __STDF__ itself (to detect "this is Fortran not C not C++") is useful.

But yes, it's a new feature, and there's an argument that it should be treated separately. I sense there is a tension here between "standardize existing behavior" and "standardize well-engineered behavior" -- and no, I don't think you can say that the current state of Fortran preprocessing is "well-engineered" by any stretch of the imagination! 😅

gklimowicz:
Well, `__STDF__` is just a new macro name that tries to standardize a test "is this Fortran?". Other disparate names mean "This is Fortran", but also indicate the specific compiler. (ChatGPT tells me of  `__GFORTRAN__`, `__FLANG__`, `__INTEL_FORTRAN`, for example.)

I see value in the analog to `__STDC__`.

gklimowicz:
With respect to pulling out some of the individual features into separate papers, I don't think I have an argument against that. If JoR wants to put forth requirements for the minimum behavior seen in existing codes, we can do that.

But as @bonachea pointed out to me once, "modern" CPP features don't appear in projects because compiler support is lacking. They'd use them if more compilers were required to supported them.

Until C23, CPP hasn't changed a lot since C99. Function-like macros with variable arguments have been standard in C for 25 years.

bonachea (changes requested):
I'm opposed to this PR, and strongly opposed to the removal of `__VA_ARGS__`. 

However I'm willing to be outvoted if there's strong pushback from opposition.

aury6623:
I'm confused on how `__STDF__` would ever be useful.

> Well, __STDF__ is just a new macro name that tries to standardize a test "is this Fortran?"

We know it's Fortran because the code is in a Fortran source file. Right? When would `__STDF__ ` ever be undef except with a FPP that hasn't implemented it yet?

All that `__STDF__` says it that your FPP decided to implement that one extremely low hanging fruit of a standard FPP feature. It won't tell you anything else about what features the FPP might support. Meanwhile the compiler-specific things like `__FLANG__` tell you a lot more useful information.

aury6623:
> but that __STDF__ itself (to detect "this is Fortran not C not C++") is useful.

Are there cases where there are header files that contain both C and Fortran code and the preprocessor selects between them? (And if so... why are they not just separate files?) (And furthermore, you could just check `__STDC__` instead!)

bonachea:
> Are there cases where there are header files that contain both C and Fortran code and the preprocessor selects between them?

Yes. This is the motivation literally explained in the text that your PR is proposing to delete.

It's the same reason why both `__STDC__` and `__cplusplus` exist to denote the language, even though technically only one is necessary for mixed-language headers files IFF the universe of possible languages is "closed" (which it's clearly not, Fortran being the immediate third example).

aury6623:
> Yes. This is the motivation literally explained in the text that your PR is proposing to delete.

My question was whether or not those cases already exist; the text in the PR makes it sound like a hypothetical.

I don't understand the purpose of writing a mixed language header file. C and C++ mixed language makes sense because a header file for the both could be nearly identical. For Fortran and C, the contents would be entirely different. Why not just make them be multiple files? Why are we encouraging the use of _more_ preprocessor shenanigans?

bonachea:
> I don't understand the purpose of writing a mixed language header file with Fortran. C and C++ mixed language makes sense because a header file for the both could be nearly identical. For Fortran and C, the contents would be entirely different.

One immediate example that comes to mind is a header file containing almost exclusively preprocessor directives, with possibly just a few language-specific declarations. The config.h file auto-generated by the popular autoconf tool is one example of that.

PR #28, opened 2025-02-06
Add paper as uploaded to the J3 web site as 25-114, and rename

gklimowicz:


gklimowicz:
Merging to checkpoint before starting on r1.

PR #30, opened 2025-02-07
Update with the changes made by uploading 25-114.txt to J3 website

gklimowicz:
We probably should have done this sooner.
I too frequently forget about the upload changes.
This should be rolled into the changes for a 25-114r1.

### DIFF
--- a/drafts/25-114.txt
+++ b/drafts/25-114.txt
@@ -1,5 +1,5 @@
 To: J3                                                     J3/25-114
-From: Gary Klimowicz, Dan Bonachea, Aury Shafran
+From: Gary Klimowicz & Dan Bonachea & Aury Shafran
 Subject: Fortran preprocessor requirements
 Date: 2025-February-06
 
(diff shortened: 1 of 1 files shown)


gklimowicz:
There is no rush on this, but it's something that should be done before uploading `25-114r1.txt`. Otherwise it will get a new paper number, which we don't want.

drafts/25-114.txt
```diff
@@ -1,7 +1,7 @@
-To: J3                                                     J3/25-xxx
+To: J3                                                     J3/25-114
 From: Gary Klimowicz, Dan Bonachea, Aury Shafran
```
bonachea, line 2: Please also fix the author formatting to ensure the system correctly detects the author list.

IIUC this needs to be:
```suggestion
From: Gary Klimowicz & Dan Bonachea & Aury Shafran
```

bonachea (changes requested):
Missed one detail, otherwise LGTM

drafts/25-114.txt
gklimowicz: Change applied in commit 72081ef.

bonachea (approved):

aury6623 (approved):

PR #32, opened 2025-02-07
Add draft Bison grammar based on paper 25-114 (r0)

gklimowicz:
This is a draft Bison grammar for the proposed Fortran preprocessor. It specifies no semantics and relies on cleverness in the lexical analyzer to manage the
context-sensitive aspects of whitespace in the preprocessor.

This reflects the grammar proposed it 25-114. There are many rules commented out from the before-times when I anticipated a richer set of operations. I didn't want to lose those yet, so they are commented out and marked as TBD.

Bison accepts this grammar and produces a valid '.tab.c' file (modulo the missing yylex() and yyerror() routines).

### DIFF
--- a/drafts/25-nnn-bison-grammar-only.y
+++ b/drafts/25-nnn-bison-grammar-only.y
@@ -1,0 +1,532 @@
+/*
+ * A Bison grammar for Phase 3 of the Fortran preprocessor.
+ *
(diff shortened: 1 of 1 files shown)


gklimowicz:
This pull request should not have boon made against the main repo. If we need it, we will reintroduce it in a forked repo.

Issue #33, opened 2025-02-09
Add all the operators that CPP allows in #if / #elif expressions


bonachea:
The C operator list in the current 25-114 draft:

> From C: && || == != < > <= >= + / * ! & | ^ ~ ( )

is missing several C operators that are permitted by the C23 grammar for _constant-expression_ and additionally permitted by C23 in `#if` and `#elif` directive expressions. These are:

1. `?:`  (_conditional-expression_, ternary operator)
2. `<<` and `>>` (_shift-expression_, bit shifting)
3. `%` (_multiplicative-expression_, integer modulus)
4. binary `-` (_additive-expression_, subtraction)
5. unary `-` (_unary-operator_, negation)

I assume this was simply an oversight as these operators are less common in preprocessor directives, as opposed to a deliberate decision to exclude them. 

Demonstration of these operators in [C23 CPP](https://godbolt.org/z/77fxdEhq1) and [gfortran 14.2 preprocess](https://godbolt.org/z/odMPGqsaG)

Here I'm deliberately excluding operators like unary `&` and `*`, _unary-expression_ (`sizeof`, `alignof`, `++`, `--`), _postfix-expression_ (array access, member access, function call) and _cast-expression_, all of which are grammatically permitted by _constant-expression_  but prohibited in integer constant expressions (additional constraint on `#if` and `#elif`) and thus irrelevant for our purposes.

bonachea:
Resolved in #34

PR #34, opened 2025-02-09
Add all the operators that CPP allows in #if / #elif expressions

bonachea:
This adds the five-ish operators accepted by CPP in `#if` and `#elif` that were missing from 25-114r0.

Also rearrange/regroup the operator list and add explanatory parentheticals, for readers less familiar with C grammar. I assume a future syntax paper will explode this out further to show the full grammar, but that seems redundant for requirements and I think this conveys the right intuition without being too pedantic.

Resolves issue #33


gklimowicz (approved):
Very nice. Thank you.

PR #37, opened 2025-02-10
No expansion: remove mention of FORMAT and IMPLICIT

gklimowicz:
After discussion, remove prohibition on expansion in `FORMAT` specifications, and `IMPLICIT` statements. Also, treat Hollerith data and format specifiers as we treat character constants.

### DIFF
--- a/drafts/25-114.txt
+++ b/drafts/25-114.txt
@@ -239,8 +239,8 @@ FPP will not expand in fixed-form
 
 FPP will not expand tokens in either fixed- or free-form:
     - In character constants
-    - In FORMAT statements
-    - In the letter-spec-list in an IMPLICIT statement.
+    - In Hollerith constants in DATA statements and argument lists,
+    - In Hollerith format descriptors.
 
 
 4.6 Output of Phase 2
--- a/drafts/25-114.txt
+++ b/drafts/25-114.txt
@@ -238,7 +238,8 @@ FPP will not expand in fixed-form
 
 
 FPP will not expand tokens in either fixed- or free-form:
-    - In character constants
+    - In character literal constants.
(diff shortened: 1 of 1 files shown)


gklimowicz:
After discussion with JoR today, we agree

- We should remove the references to `FORMAT` and `IMPLICIT`.
- We should *not* insert anything about Hollerith.

gklimowicz:
Updated PR to reflect decisions in JoR meeting today.

gklimowicz:
After discussion with JoR today, we agree

We should remove the references to FORMAT and IMPLICIT.
We should not insert anything about Hollerith.

drafts/25-114.txt
```diff
@@ -239,8 +239,6 @@ FPP will not expand in fixed-form
 
 FPP will not expand tokens in either fixed- or free-form:
     - In character constants
```
bonachea, line 241: The standard uses the phase "character literal constant" in 7.4.4.3, and I think we should echo that here.

Also do we need to also explicitly mention character string (ie quoted) edit descriptors?

IIUC those are not technically character literal constants (see 13.9), but we also want to prohibit expansion there (partially because the preprocessor lacks context information to discern whether a quoted character string token appears in an edit descriptor).

```suggestion
    - In character literal constants
    - In character string edit descriptors (quoted character strings) 
```

bonachea (changes requested):
I think we are in general agreement on content, but I have some wording quibbles and added suggestions.

drafts/25-114.txt
sblionel: I agree with Dan.

gklimowicz: I agree with Steve and Dan. Update forthcoming.

PR #40, opened 2025-02-10
Rename section 4.2 and move __VA_*

bonachea:
CPP _recognizes_ and processes certain tokens (`#`, `##`, `__VA_ARGS__`, `__VA_OPT__`, and FLM arguments), in the replacement-list of macro. But it never "rejects" anything whatsoever, any unrecognized tokens are just passed through uninterpreted to expansion. So it's not correct to talk about "accepting tokens" in that context.

`__VA_ARGS__`, `__VA_OPT__` really belong in this section instead of "4.4 Macros defined by the preprocessor" where they were misleadingly placed before, so move them.

### DIFF
--- a/drafts/25-114.txt
+++ b/drafts/25-114.txt
@@ -186,13 +186,21 @@ interpolated by INCLUDE lines will be treated as if it had been included
 via #include.
 
 
-4.2 Tokens accepted on #define directives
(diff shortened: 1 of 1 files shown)


gklimowicz (approved):
LGTM.

PR #45, opened 2025-02-13
Add tutorial Preprocessor Take 2.pdf

gklimowicz:
Though not an official J3 paper, this tutorial was instrumental in gaining consensus on the rough requirements for the preprocessor.


bonachea (changes requested):
This PDF appears to lack important metadata, such as the authorship credit, original date and venue.

Assuming we have that information, I'd like us to record it somehow.

We should at least provided it in the commit introducing the PDF, but even better would be to also record it in the PDF metadata fields intended for this purpose.

gklimowicz:
I may not have all the tooling to do that, but I will see what I can do.

bonachea:
> I may not have all the tooling to do that, but I will see what I can do. (I don't usually wrangle PDFs that much except to export to the format.)

[ExifTool](https://exiftool.org/) can do it, along with a number of other PDF tools.

I can easily do it if that's easier, but I lack the relevant information.

gklimowicz:
I need to do a new PR with the metadata update.

PR #49, opened 2025-02-19
Expand in only full-line comments; don't expand in alternative chars

gklimowicz:
We no longer try to discern "comment directives" for macro expansion. A simpler proposal to expand in (full-line) comment lines is added.

A proposal to mention no token expansion in "alternative syntax for character constants and edit descriptors" was added. (We're looking at you, Hollerith.)

Add reference for the Fortran 2023 Interpretation document, 24-007, as we use it to define "comment line" meaning a comment that contains all the non-blank portion of a line.

### DIFF
--- a/drafts/25-114.txt
+++ b/drafts/25-114.txt
@@ -6,6 +6,7 @@ Date: 2025-February-06
 References: 95-257 Conditional Compilation: The FCC Approach.txt
             96-063 A Fortran Preprocessor.txt
             23-192r1 F202Y Define a standard Fortran preprocessor.txt
+            24-007 Fortran 2023 Interpretation Document
             24-108 Preprocessor directives seen in existing Fortran
                    programs.txt
             24-109 On Fortran awareness in a Fortran preprocessor.txt
@@ -138,8 +139,8 @@ Comment handling:
     3. FPP recognizes /* ... */ C-style comments on directive lines.
        /* ... */ comments are not recognized in (non-directive)
        Fortran source lines.
-    4. Macros are expanded in comment lines that appear to be directives
-       (processor-specific, such as '!$omp', or '!$acc', or others).
+    4. Macros are expanded in comment lines as described in clauses
+       6.3.2.3 and 6.3.3.2 in the Fortran 2023 interpretation document.
 
 Constant expressions in #if and #elif:
     1. Expressions in #if and #elif directives allow most operators
@@ -245,6 +246,9 @@ FPP will not expand in fixed-form:
 FPP will not expand tokens in either fixed- or free-form:
     - In character literal constants.
     - In character string edit descriptors (quoted character strings).
+
+FPP will not expand tokens in any alternative syntax for character
+literal constants or edit descriptors that the processor supports.
 
 
 4.6 Output of Phase 2
--- a/drafts/25-114.txt
+++ b/drafts/25-114.txt
@@ -139,8 +139,11 @@ Comment handling:
     3. FPP recognizes /* ... */ C-style comments on directive lines.
        /* ... */ comments are not recognized in (non-directive)
        Fortran source lines.
-    4. Macros are expanded in comment lines as described in clauses
-       6.3.2.3 and 6.3.3.2 in the Fortran 2023 interpretation document.
+    4. On (non-directive) Fortran source lines, macros are expanded within
+       all Fortran-style comment lines (i.e., as defined by clauses 6.3.2.3
+       and 6.3.3.2 in the Fortran 2023 interpretation document). This notably
+       implies macro expansion within processor-specific "directive-style"
+       comment lines (such as '!$omp', or '!$acc', or others).
 
 Constant expressions in #if and #elif:
     1. Expressions in #if and #elif directives allow most operators
@@ -247,8 +250,10 @@ FPP will not expand tokens in either fixed- or free-form:
     - In character literal constants.
     - In character string edit descriptors (quoted character strings).
 
-FPP will not expand tokens in any alternative syntax for character
-literal constants or edit descriptors that the processor supports.
+If alternative syntax for character literal constants is available,
+FPP will not expand tokens in either fixed- or free-form:
+    - In the alternative syntax for character literal constants.
+    - In the alternative syntax for character string edit descriptors.
 
 
 4.6 Output of Phase 2


drafts/25-114.txt
```diff
@@ -138,8 +139,8 @@ Comment handling:
     3. FPP recognizes /* ... */ C-style comments on directive lines.
        /* ... */ comments are not recognized in (non-directive)
        Fortran source lines.
-    4. Macros are expanded in comment lines that appear to be directives
-       (processor-specific, such as '!$omp', or '!$acc', or others).
+    4. Macros are expanded in comment lines as described in clauses
+       6.3.2.3 and 6.3.3.2 in the Fortran 2023 interpretation document.
```
bonachea, line 143: **Semantics:**    
Is it really our intent to restrict this to "comment lines", ie "If the first nonblank character on a line is an `!`, the line is a comment line." Or are we stating a rule for ALL Fortran comments, regardless of whether `!` appears at the start or in the middle of a physical line? 

I honestly don't know because I didn't realize this subtlety was in-play during the discussion. Also I'm not sure if "directive-style" comments ever appear on line with code or if they are always on a line alone?

**Presentation:** 
As phrased this wording could be misread to mean that macro expansion is defined in those specification sections, when our intent is to define the term "comment lines". 

Also given the previous rule is discussing directive lines, I think we should explicitly clarify this applies to comments on non-directive lines.

Finally I think it's worth explicitly emphasizing the point of this requirement, which is to allow expansion impact processor-specific "directive-style" comments.

Assuming we want ALL comments, something like:
```suggestion
    4. On (non-directive) Fortran source lines, macros are expanded within all 
       commentary text (i.e., within all Fortran-style comments). 
       This notably implies macro expansion within processor-specific 
       "directive-style" comments (such as '!$omp', or '!$acc', or others).
```

Assuming we want just comment lines, perhaps:
```suggestion
    4. On (non-directive) Fortran source lines, macros are expanded within all 
       comment lines (i.e., within Fortran-style full-line comments, as defined by 
       clauses 6.3.2.3 and 6.3.3.2 in the Fortran 2023 interpretation document). 
       This notably implies macro expansion within processor-specific 
       "directive-style" comment lines (such as '!$omp', or '!$acc', or others).
```

drafts/25-114.txt
```diff
@@ -246,6 +247,9 @@ FPP will not expand tokens in either fixed- or free-form:
     - In character literal constants.
     - In character string edit descriptors (quoted character strings).
 
+FPP will not expand tokens in any alternative syntax for character
+literal constants or edit descriptors that the processor supports.
```
bonachea, line 251: For the record, Malcolm's literal wording suggestion here was: (I took careful notes)
> “if a processor has an alternative syntax for character literal constants, macro expansion shall not take place within that literal constant”

I'm not opposed to massaging that wording, but I worry about the potential for unintended misreadings. For example, the text above could be misparsed/misread as "FPP will not expand tokens in ... edit descriptors that the processor supports" which is too broad and definitely not what we mean.

Do we really need to mention edit descriptors explicitly? Doesn't a Hollerith constant appearing in an edit descriptor still qualify as an "character literal constant (with alternate syntax)"?

```suggestion
literal constants that the processor supports.
```

bonachea (changes requested):
I have questions for discussion, and some potential resolutions.

Let's iterate in comment threads.

drafts/25-114.txt
```diff
@@ -6,6 +6,7 @@ Date: 2025-February-06
 References: 95-257 Conditional Compilation: The FCC Approach.txt
             96-063 A Fortran Preprocessor.txt
             23-192r1 F202Y Define a standard Fortran preprocessor.txt
+            24-007 Fortran 2023 Interpretation Document
```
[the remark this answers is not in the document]
gklimowicz: The convention is to reference the most current "interpretation document" for the most recent release. 24-007 was the interp doc that Malcolm created after the release of F2023. He has a new one, 25-007, but that is his first attempt to update fonts and LaTeX engine for ISO requirements. It has issues (can't search ligatures, etc.). So I used the last good one created under the old tools. Once 25-007 is decent, we can use that.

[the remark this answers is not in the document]
gklimowicz: When we get down to edits, we'll probably be looking at that newest draft (prob 26-007)

drafts/25-114.txt
gklimowicz: I was trying to mirror the text that preceded it for character literal constants and edit descriptors. The standard treats these as two different things, and calls them out separately. Now, the meaning is as if I could parenthesize this as "(character literal) constants or edit descriptors". The proper form is probably to hyphenate the compound adjective character-literal constants or edit descriptors.

drafts/25-114.txt
gklimowicz: The standard doesn't talk about full-line comments. The term used is "comment lines", so that's what I used. It's a term only used in those two subclauses and a neighboring clause or two. With Malcolm, it's usually best to the terminology the standard uses. I thought about putting something like "'full' comment lines" but Malcolm will surely jump on that: "If you mean 'comment line', say it."

drafts/25-114.txt
bonachea: The potential misparse is to drop the entire phrase "alternative syntax for character literal constants" before the "or" and misread the second half as a prohibition on expansion in "any .. edit descriptors". I don't see how adding parens or hypens to the dropped text prevents that misreading.

My point with the suggestion (and my understanding of Malcolm's suggested wording) is this rule applies to Hollerith literal constants (in code) and Hollerith literal constants appearing inside an edit descriptor, where I _assume_ the latter still qualify as "literal constants"... but I don't have a copy of the ancient Fortran spec that supported these to verify that reading.

drafts/25-114.txt
bonachea: Let's resolve semantics before wording.

Based on your reply, I assume you are advocating for the rule semantics to apply ONLY to "comment lines" (a.k.a full line comments), and NOT to comment text that starts on a line after some code? 

How confident are we that such "partial line" comments never contain processor-specific "directive-style" comments, or that we can safely ban expansion there?

gklimowicz: Yes, semantics only for comment lines (a.k.a. full-line comments) is my intent, and I believe Malcolm's complaint as well. I am not _entirely_ confident that there are no directives in lines that also contain Fortran code. (Especially true for the menagerie of legacy DEC/Intel/Cray/GFortran comment directives.) OpenMP requires that the comment directives in Fortran be the first non-blank character in free-form and full-line comments in fixed-form (Open MP 5.0, sections 2.1.1 and 2.1.2). OpenACC is the same (I only have OpenACC 3.3; it's in section 2.1).

gklimowicz: So I now feel much more confident that the _standard_ should say "comment lines", and leave it to the people who support the menagerie of legacy stuff (`!dir$`, etc.) to support them as they require.

bonachea: Ok, updated proposal based on discussion:

```suggestion
    4. On (non-directive) Fortran source lines, macros are expanded within all 
       Fortran-style comment lines (i.e., as defined by 
       clauses 6.3.2.3 and 6.3.3.2 in the Fortran 2023 interpretation document). 
       This notably implies macro expansion within processor-specific 
       "directive-style" comment lines (such as '!$omp', or '!$acc', or others).
```

drafts/25-114.txt
bonachea, line 251: If we cannot resolve the above in short order, then in the interests of time I'll suggest reverting to something based on the phrasing Malcolm provided:

```suggestion
If a processor provides an alternative syntax for character literal constants, 
FPP macro expansion shall not take place within such literal constants.
```

If someone very picky wants to stand up and scream that F66 Hollerith edit descriptors did not technically qualify as "literal constants", then we can try to reach a wording they'll agree to for the syntax paper, but this doesn't seem critical to wordsmith for requirements.

gklimowicz: Yeah, the assumption is not correct. They've been called out separately as "edit descriptors" since Fortran 77. (In Fortran 66 they were called "field descriptors".) The modern standard has them as such in subclause 13.3.2. Now, it _does_ say in R1322 that a _char-string-edit-desc_ is  _char-literal-constant_. But they are semantically distinct concepts still in the standard, even if they are syntactically identical. So I want to leave the distinction.

gklimowicz: When he brought up Hollerith, I took that as a sign that he was having to really reach to find stuff to complain about.

bonachea: We could possibly resolve the potential misparse and explcitly include edit descriptors by pedantically breaking it into two sentences:

```suggestion
literal constants that the processor supports.
FPP will not expand tokens in any alternative syntax for character
string edit descriptors that the processor supports.
```

gklimowicz: I don't want to set us up to have to do a 25-114r3

bonachea: Relevant text from B.1-5:

> H edit descriptor.
> In Fortran 77 and Fortran 90, there was an alternative form of character string edit descriptor,
> which had been the only such form in Fortran 66; this has been deleted. A similar result can be
> achieved by using a character string edit descriptor.

drafts/25-114.txt
gklimowicz: I'm good, subject to line length checks.

drafts/25-114.txt
gklimowicz: Yes. This is why Steve didn't want to mention it at all. It's been deleted for 30 years. (Taken to the extreme, need we mention that we will do macro expansion in arithmetic `IF`?  `PAUSE` statements? `ASSIGN` and assigned `GO TO` statements?)

gklimowicz:
Want me to take a stab now at an update to the branch? (I fear I will botch the branch push...)

drafts/25-114.txt
bonachea: I agree it's a bit tangential, but the consensus from discussion was for us to specifically mention this one obsolescent feature.

Your thoughts on my latest proposal suggestion above?

bonachea:
> Want me to take a stab now at an update to the branch? (I fear I will botch the branch push...)

Yes if you have new commits to push to the branch, please do so. If you botch it somehow I can help you rewrite history to fix it.

In any case please do NOT close this PR and open a new one, which should never be done after review has begun; I want to preserve our discussion.

bonachea (approved):
LGTM

PR #52, opened 2025-03-29
Update section 2 with several additions for review

gklimowicz:
- Renumber spec lists a bit to get unique numbers and allow for insert additional specifications a bit more easily.

- Collapse original section 2 "Formal specifications" into its main "Lexical specifications" to include lines, whitespace, etc.

- Add Patrick Fasano comments on multiple whitespace characters (but be explicit about it being in non-character-constant contexts).

- Remove the over-specified list of tokens. This is more appropriate for a syntax paper. Introduce tokens enough to allow us to use them in illustrative syntax. (We can remove the explicitness about the tokens `ID`, `WHOLE_NUMBER`, `REAL_NUMBER`, `CHARACTER_STRING`, and `UNRECOGNIZED_TOKEN` if we like.)

### DIFF
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -59,9 +59,8 @@ li17. Preprocessor directive continuation processing described by the prior rule
 li19. The maximum length of a preprocessor directive (including
       continuation text) is one million characters.
 
-li21. A non-empty source file shall not end with a directive line
-      that ends with a '\' followed
-      immediately by a new-line (as required in C 2023 5.1.1.2 bullet 2).
+li21. A source file that ends with a directive line shall neither end with a '\', 
+      nor a '\' followed immediately by a new-line (analogously to C 2023 5.1.1.2 bullet 2).
 
 li31. Fortran comment lines are defined as in 25-007 6.2.2.3 and
       6.2.3.2.
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -62,8 +62,8 @@ li19. The maximum length of a preprocessor directive (including
 li21. A source file that ends with a directive line shall neither end with a '\', 
       nor a '\' followed immediately by a new-line (analogously to C 2023 5.1.1.2 bullet 2).
 
-li31. Fortran comment lines are defined as in 25-007 6.2.2.3 and
-      6.2.3.2.
+li31. Fortran comment lines are defined as in 25-007 6.3.2.3 and
+      6.3.3.2.
 
 li41. Fortran source fragments are those lines that are neither
       preprocessor directive lines (or continuations thereof)
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -28,53 +28,52 @@ discussions, these have included line continuation processing, comment
 handling, and tokenization for the preprocessor.)
 
 
-2. Formal specifications
-========================
+2 Lexical specifications
+------------------------
 
-2.1 Lexical specifications
---------------------------
-
-2.1.1 Lines
------------
+2.1 Lines
+---------
 
 li00. The Fortran preprocessor recognizes three distinct types of lines:
       preprocessing directives (and continuation lines thereof),
       Fortran comment lines, and Fortran source fragments.
 
-li10. A line that has a '#'
+li11. A line that has a '#'
       character as the first non-blank character of the line is a
       directive line (as required by C 2023 6.10.1 paragraph 2),
       except when otherwise specified by the next two rules.
 
-li21. In fixed-form input files, a '#' in column 6 of a non-comment
+li13. In fixed-form input files, a '#' in column 6 of a non-comment
       line does not introduce a directive line.
 
-li22. A preprocessor directive can be continued with a backslash '\'
+li15. A preprocessor directive can be continued with a backslash '\'
       immediately followed by a new-line. The backslash and new-line are
       deleted, the content of the subsequent line are textually appended
       to the current directive, and the subsequent line is deleted.
       This process repeats until the current directive does not end
       with a backslash '\' immediately followed by a new-line.
 
-li23. Preprocessor directive continuation processing described by the prior rule is
+li17. Preprocessor directive continuation processing described by the prior rule is
       effectively performed before any other processing of the text in affected lines.
 
-li23. The maximum length of a preprocessor directive (including
+li19. The maximum length of a preprocessor directive (including
       continuation text) is one million characters.
 
-li24. A non-empty source file shall not end with a '\' followed
-      immediately by a new-line as required in C 2023 5.1.1.2 bullet 2.
+li21. A non-empty source file shall not end with a directive line
+      that ends with a '\' followed
+      immediately by a new-line (as required in C 2023 5.1.1.2 bullet 2).
 
-li30. Fortran source fragments are those lines that are neither
+li31. Fortran comment lines are defined as in 25-007 6.2.2.3 and
+      6.2.3.2.
+
+li41. Fortran source fragments are those lines that are neither
       preprocessor directive lines (or continuations thereof)
       nor Fortran comment lines.
 
-
-
-li31. Text on fixed-form Fortran source fragments is ignored beyond
+li43. Text on fixed-form Fortran source fragments is ignored beyond
       column 72.
 
-li32. Fortran source fragments may be continued with a continuation
+li45. Fortran source fragments may be continued with a continuation
       ('&' at the end of a free-form line as specified in Fortran 2023
       6.3.2.4, or with a non-blank, non-zero character in column 6
       of a fixed-form line as specified in Fortran 2023 6.3.3.3).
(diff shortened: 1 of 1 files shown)


sblionel (approved):

kc9jud (approved):

JonSteidel (approved):

drafts/25-xxx-specifications.txt
```diff
@@ -28,53 +28,52 @@ discussions, these have included line continuation processing, comment
...
       continuation text) is one million characters.
 
-li24. A non-empty source file shall not end with a '\' followed
-      immediately by a new-line as required in C 2023 5.1.1.2 bullet 2.
+li21. A non-empty source file shall not end with a directive line
+      that ends with a '\' followed
+      immediately by a new-line (as required in C 2023 5.1.1.2 bullet 2).
+
+li31. Fortran comment lines are defined as in 25-007 6.2.2.3 and
+      6.2.3.2.
```
bonachea, line 67: The referenced sections don't exist.

I think you mean 6.3.2.3 and 6.3.3.2

drafts/25-xxx-specifications.txt
```diff
@@ -99,68 +98,107 @@ li32. Fortran source fragments may be continued with a continuation
...
+and the execution behavior of the preprocess on Fortran comment lines
+and Fortran source fragment lines.
+
+This illustrative syntax makes use of these "tokens".
+Further definition of the recognized tokens is deferred to the upcoming preprocessor syntax paper.
+
+
+to01. In definitions of object macros and function/like macros,
+      the tokens that may appear in a replacement list include those
+      allowed in Fortran and those allowed in C integer expressions.
```
bonachea, line 177: As written, this rule could easily be misread as overly constraining. 

In CPP, a `#define` replacement list may contain literally any text (other than a newline). This includes things like C keywords, or even unknown tokens (e.g. `#$&%`).

The rule is technically accurate because indeed the "tokens .. may .. include" the subset you name, but they also may include plenty of other tokens outside the named subset. IMO we should either name the complete permitted set (literally any text other than newline), or use the words "for example" to clarify we are naming a proper subset.

drafts/25-xxx-specifications.txt
```diff
@@ -99,68 +98,107 @@ li32. Fortran source fragments may be continued with a continuation
...
+
+to03. Without naming them all explicitly, they appear in the illustrative
+      syntax in subsequent sections.
```
bonachea, line 180: Unclear what "them" and "they" refer to here.

I think we generally want each rule to stand-alone grammatically, and possibly refer explicitly to other rules when necessary.

drafts/25-xxx-specifications.txt
```diff
@@ -99,68 +98,107 @@ li32. Fortran source fragments may be continued with a continuation
...
+
+to05. Directive names are recognized at the beginning of a line.
```
bonachea, line 182: No idea what this rule is trying to say.

Section 2.1 already defined "directive line", but the term "Directive names" has not been defined.

drafts/25-xxx-specifications.txt
```diff
@@ -99,68 +98,107 @@ li32. Fortran source fragments may be continued with a continuation
...
+
+to07. The '#' character on a directive line may be preceded or followed
+      by whitespace.
+
+to09. The preprocessor is line oriented. To define the end of
```
bonachea, line 187:
```suggestion
to09. The preprocessor is line-oriented. To define the end of
```

drafts/25-xxx-specifications.txt
```diff
@@ -28,53 +28,52 @@ discussions, these have included line continuation processing, comment
...
+---------
 
 li00. The Fortran preprocessor recognizes three distinct types of lines:
       preprocessing directives (and continuation lines thereof),
       Fortran comment lines, and Fortran source fragments.
 
-li10. A line that has a '#'
+li11. A line that has a '#'
       character as the first non-blank character of the line is a
       directive line (as required by C 2023 6.10.1 paragraph 2),
```
bonachea, line 43: We are using `C 2023` throughout to refer to the C spec, but the listing in the references is:

```
ISO/IEC 9899:2023 Programming languages -- C
                 (working draft N3096)
```

I see two options:
1. Replace all citations with `ISO/IEC 9899:2023`, or
2. Change the references to introduce a "short form", perhaps something like:

```
ISO/IEC 9899:2023 Programming languages -- C ("C 2023")
                 (working draft N3096)
```

drafts/25-xxx-specifications.txt
```diff
@@ -99,68 +98,107 @@ li32. Fortran source fragments may be continued with a continuation
...
 
 co01. Directive lines may contain C-style '/*' ... '*/' comments.
 
-co02. Directive lines shall not contain C style '//' comments.
+co03. Directive lines shall not contain C style '//' comments.
 
-co02. '/*' ... '*/' comments on directive lines shall extend past a
+co05. '/*' ... '*/' comments on directive lines shall extend past a
       new-line only if the line ends in '\' new-line, indicating a
       continuation line. (Maybe controversial)
```
bonachea, line 146: I think we can remove the non-normative commentary here.

Everything is potentially controversial.

```suggestion
      continuation line.
```

drafts/25-xxx-specifications.txt
```diff
@@ -99,68 +98,107 @@ li32. Fortran source fragments may be continued with a continuation
...
-constants, the Fortran operators, and the C operators used to construct
-constant-integer-expression.]
+
+The preprocessor is a token-manipulation engine (See C 2023 5.1.1.2
+Translation phases).
+As such, there is a specific lexicon of tokens recognized by
+the preprocessor (including unrecognizable tokens).
+
+We use illustrative syntax to describe the directive specifications,
+and the execution behavior of the preprocess on Fortran comment lines
```
bonachea, line 168: I continue to strongly oppose misusing the term "execution" to refer to any part of compilation.

I believe the C spec exclusively uses the term "execution" to describe program execution. Any actions taken by the preprocessor or compiler are "translation" (see C 23 5.0-1). Based on a quick skim, the same appears to be true of the Fortran spec.

```suggestion
and the translation behavior of the preprocess on Fortran comment lines
```

bonachea (changes requested):
Thanks for the continuing work on this!

I added several comments that I'd like to see resolved in the near-term. 

A few of my comments (those regarding existing text) are somewhat orthogonal to this PR, and those could potentially be moved to the issue tracker if you'd rather handle them outside the scope of this PR.

drafts/25-xxx-specifications.txt
gklimowicz: I believe the technical description for this error is "dammit!"

I don't know how to suggest my own changes to the PR, so feel free to suggest the correction and I will be happy to commit it.

If not, I can update the PR. (I want to fix all your comments...)

bonachea:
```suggestion
li31. Fortran comment lines are defined as in 25-007 6.3.2.3 and
      6.3.3.2.
```

bonachea (commented):
Thanks for the updates, a number of comment threads requesting changes remain open.

drafts/25-xxx-specifications.txt
gklimowicz: I think "to05" should be moved to the section on whitespace. Im adding a "ws02".

drafts/25-xxx-specifications.txt
gklimowicz: I'll update this PR.

gklimowicz: OK, I think I have updated the PR, but I am easily confused.

drafts/25-xxx-specifications.txt
gklimowicz: I have updated the PR with more liberal wording. You are sure to have improvements.

drafts/25-xxx-specifications.txt
gklimowicz: Updated PR to be explicit what we are talking about. Moved to05 to the "whitespace" section.

drafts/25-xxx-specifications.txt
```diff
@@ -99,68 +97,109 @@ li32. Fortran source fragments may be continued with a continuation
...
+to01. In the definitions of object macros and function-like macros,
+      the tokens that may appear in a replacement list include those
+      allowed in Fortran source, those allowed in C integer expressions,
+      and any additional tokens recognized by the processor.
+
+to03. Without naming all the tokens explicitly, they appear in the illustrative
+      syntax in subsequent sections.
+
+to07. The '#' character on a directive line may be preceded or followed
+      by whitespace.
```
bonachea, line 186: This is ambiguously worded, and I believe it's now entirely duplicative of ws02

```suggestion
```

drafts/25-xxx-specifications.txt
```diff
@@ -99,68 +97,109 @@ li32. Fortran source fragments may be continued with a continuation
...
+and Fortran source fragment lines.
+
+This illustrative syntax makes use of these "tokens".
+Further definition of the recognized tokens is deferred to the upcoming preprocessor syntax paper.
+
+
+to01. In the definitions of object macros and function-like macros,
+      the tokens that may appear in a replacement list include those
+      allowed in Fortran source, those allowed in C integer expressions,
+      and any additional tokens recognized by the processor.
```
bonachea, line 180: Given that "any additional tokens recognized by the processor" includes "unrecognized token", 
this rule really boils down to approximately mean: "the replacement list may include any arbitrary sequence of characters that doesn't include a new-line". 

I think it would be clearer to just state that directly, how about:

```suggestion
to01. In the definitions of object macros and function-like macros,
      the replacement list may include any arbitrary sequence of 
      characters that doesn't include a new-line. Once tokenized,
      this for example may include any tokens allowed in Fortran 
      source lines, those allowed in C integer expressions,
      and any additional tokens recognized by the processor.
```

bonachea (changes requested):
Just a few remaining quibbles and this should be ready to merge

gklimowicz:
Thanks for the updates.

PR #53, opened 2025-04-08
Add specifications for predefined macros

bonachea:
These are heavily based on specification prose in C23, with adjustments to match the terminology of the Fortran spec and this draft.


gklimowicz (approved):
Thank you for this. I don't see any changes I would make at this time.

PR #55, opened 2025-04-10
macros: Reserve (some) macros beginning with underscore to the implementation

### LINKED ISSUE #54: Reserving macro names
Following #53, I think the next question is whether we should go further and reserve all macros which begin with `_` for the implementation. Right now we specify that if a macro is defined by an implementation, it must start with an underscore: https://github.com/j3-fortran/fpp-202y/blob/79b75b21c9c47a6d344c82a93dff65336070e926/drafts/25-xxx-specifications.txt#L299-L301 Do we want to go further and say that only the implementation is allowed to define macros beginning with an underscore? C and C++ are quite aggressive about this (https://devblogs.microsoft.com/oldnewthing/20230109-00/?p=107685) and reserve identifiers which start with an underscore in all contexts, not just macros. Fortran already requires that names begin with a letter, so starting a name with an underscore is already forbidden, and as of yet macros are entirely unspecified.

kc9jud:
Resolves #54.

### DIFF
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -296,6 +296,13 @@ di07. The non-directive directive
 7 Predefined macros
 ===================
 
+pm00. Macro names beginning with a leading underscore followed by an uppercase
+      letter or a second underscore are reserved to the processor. Reserved
+      names shall not be the subject of a #define or #undef preprocessing
+      directive within a program unit.
+      [Note: Any macro name matching the regular expression `/^_[A-Z_][A-Za-z0-9_]*$/`
+      is considered reserved.]
+
 pm01. Any macro name predefined by the implementation shall begin with a
       leading underscore followed by an uppercase letter or a second
       underscore.
@@ -307,8 +314,7 @@ pm10. The values of the predefined macros listed in the following subclauses
       (except for `__FILE__` and `__LINE__`) remain constant throughout the
       program unit.
 
-pm12. None of the predefined macros listed in the following subclauses nor
-      the identifier `defined` shall be the subject of a #define or a #undef
+pm12. The identifier `defined` shall not be the subject of a #define or a #undef
       preprocessing directive.
 
 pm15. The presumed source file name and line number can be changed by the #line
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -303,11 +303,11 @@ pm00. Macro names beginning with a leading underscore followed by an uppercase
       [Note: Any macro name matching the regular expression `/^_[A-Z_][A-Za-z0-9_]*$/`
       is considered reserved.]
 
-pm01. Any macro name predefined by the implementation shall begin with a
+pm01. Any macro name predefined by the processor shall begin with a
       leading underscore followed by an uppercase letter or a second
       underscore.
 
-pm02. The implementation shall not predefine the macro `__cplusplus`,
+pm02. The processor shall not predefine the macro `__cplusplus`,
       nor any macro whose name starts with `__STDC`.
 
 pm10. The values of the predefined macros listed in the following subclauses
@@ -320,7 +320,7 @@ pm12. The identifier `defined` shall not be the subject of a #define or a #undef
 pm15. The presumed source file name and line number can be changed by the #line
       directive.
 
-The following macro names shall be defined by the implementation:
+The following macro names shall be defined by the processor:
 
 7.1 __LINE__
 ------------
@@ -346,7 +346,7 @@ pm41. `__DATE__` shall be a character literal constant of the form "Mmm dd yyyy"
       asctime function, and the first character of dd is a space character if the
       value is less than 10.
 
-pm42. If the date of translation is not available, an implementation-defined
+pm42. If the date of translation is not available, a processor-dependent
       valid date shall be supplied.
 
 7.4 __TIME__
@@ -359,7 +359,7 @@ pm51. `__TIME__` shall be a character literal constant of the form "hh:mm:ss",
       where hh is the hour of the day, mm is the minutes of the hour, and ss is the
       seconds of the minute.
 
-pm52. If the time of translation is not available, an implementation-defined
+pm52. If the time of translation is not available, a processor-dependent
       valid time shall be supplied.
 
 7.5 __STDF__


drafts/25-xxx-specifications.txt
```diff
@@ -296,6 +296,11 @@ di07. The non-directive directive
 7 Predefined macros
 ===================
 
+pm00. Macros beginning with a leading underscore followed by an uppercase letter
```
bonachea, line 299:
```suggestion
pm00. Macro names beginning with a leading underscore followed by an uppercase letter
```

drafts/25-xxx-specifications.txt
```diff
@@ -296,6 +296,11 @@ di07. The non-directive directive
...
+      or a second underscore are reserved to the implementation. A program unit
+      shall not define a macro beginning with an underscore followed by an
+      uppercase letter or a second underscore.
```
bonachea, line 302: This is repetitive with the previous sentence, and does not cover `#undef`. 
Finally, "the implementation" does not appear to be formally defined in Fortran. I think we mean "the processor".

How about something more like:

```suggestion
      or a second underscore are reserved to the processor. Such reserved
      names shall not be the subject of a #define or #undef preprocessing directive.
```

bonachea: With this change, this rule becomes partially redundant with pm12.

We should probably strip down pm12 to only cover the `defined` identifier

bonachea (changes requested):
As expressed in #53, I'm not super-enthusiastic about reserving large swaths of names for a vague entity.

However it's unfortunately standard practice in C & C++, and we should probably follow suit.

drafts/25-xxx-specifications.txt
```diff
@@ -296,19 +296,25 @@ di07. The non-directive directive
...
+      directive within a program unit.
+      [Note: Any valid macro name matching the regular expression `/^_[A-Z_].*/`
+      is considered reserved.]
+
 pm01. Any macro name predefined by the implementation shall begin with a
       leading underscore followed by an uppercase letter or a second
       underscore.
 
-pm02. The implementation shall not predefine the macro `__cplusplus`,
+pm02. The processor shall not predefine the macro `__cplusplus`,
```
bonachea, line 310: I think we've decided to use Fortran terminology ("the processor", "processor-dependent") instead of C terminology ("the implementation", "implementation-defined").

As such, we should probably apply that rule consistently throughout the document. I believe all the remaining instances appear below in this section.

SIDE-NOTE: C makes a subtle distinction between "unspecified behavior" and "implementation-defined behavior". Fortran seems to generally lump both categories into  "processor-dependent", so we should probably follow that convention.

bonachea (changes requested):
Thanks for the fixes.

Just one more change request to establish/restore consistency of terminology.

drafts/25-xxx-specifications.txt
kc9jud: And of course C++ actually has further distinctions between [implementation-defined behavior](https://eel.is/c++draft/intro.defs#defns.impl.defined), [unspecified behavior](https://eel.is/c++draft/intro.defs#defns.unspecified), and [undefined behavior](https://eel.is/c++draft/intro.defs#defns.undefined).

bonachea: > C++ actually has further distinctions between [implementation-defined behavior](https://eel.is/c++draft/intro.defs#defns.impl.defined), [unspecified behavior](https://eel.is/c++draft/intro.defs#defns.unspecified), and [undefined behavior](https://eel.is/c++draft/intro.defs#defns.undefined).

Well sure, all three languages have a concept of undefined behavior, but that's not relevant here.

My point was just that Fortran lacks the subtle distinction between "unspecified"("processor dependent") and "implementation-defined" (no Fortran equivalent)

bonachea:
@kc9jud IMO it's generally impolite to force push rewritten commits while a PR is actively under review. Doing so means the reviewers have to start their review over again from scratch, rather than just reviewing the fixup commits you added to address feedback, relative to their last review.

This PR is very small so it's not a big deal in this case, but it's a practice I think we should discourage.

If you're really worried about the git history (which I don't think we need to be in this repo) you always have the option to squash fixup commits as the very last step before merging a PR, once the content changes are Approved.

drafts/25-xxx-specifications.txt
bonachea, line 1: You missed one instance in converting "implementation" to "processor":

> The following macro names shall be defined by the implementation:

bonachea (changes requested):
Almost there!

drafts/25-xxx-specifications.txt
```diff
@@ -296,25 +296,31 @@ di07. The non-directive directive
 7 Predefined macros
 ===================
 
-pm01. Any macro name predefined by the implementation shall begin with a
+pm00. Macro names beginning with a leading underscore followed by an uppercase
+      letter or a second underscore are reserved to the processor. Reserved
+      names shall not be the subject of a #define or #undef preprocessing
+      directive within a program unit.
+      [Note: Any valid macro name matching the regular expression `/^_[A-Z_].*/`
```
gklimowicz, line 303: Two things: I would rather we call out the full regular expression, not leave it to the reader to piece together `valid macro name` and the use of `.*`.

Also, since `^` means beginning-of-line in regex-speak, I would prefer to omit it. We're matching a macro name, not a line.

```suggestion
      [Note: Any macro name matching the regular expression `/_[A-Z][A-Za-z0-9_]*/`
```

gklimowicz (changes requested):
Just a nitpick about the regular expression in the note for pm00.

drafts/25-xxx-specifications.txt
kc9jud: @gklimowicz Explicit is good, agreed on expanding out the macro.

Re: `^`, according to POSIX anyway, [`^` anchors the expression to the beginning of the string](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap09.html#tag_09_03_08) rather than the beginning of the line. (To have `^` match the beginning of any line, you have to add the `m` modifier in _e.g._ PCRE.)

bonachea:
> @kc9jud IMO it's generally impolite to force push rewritten commits while a PR is actively under review. Doing so means the reviewers potentially have to start their review over again from scratch, rather than just reviewing the fixup commits you added to address feedback, relative to their last review.
> 
> This PR is very small so it's not a big deal in this case, but it's a practice I think we should discourage.
> 
> If you're really worried about the git history (which I don't think we need to be in this repo) you always have the option to squash fixup commits as the very last step before merging a PR, once the content changes are Approved.

@kc9jud ping again, please stop force-pushing to rewrite reviewed commits.

drafts/25-xxx-specifications.txt
kc9jud: by the same argument, I guess the "correct" regex is actually `/_[A-Z][A-Za-z0-9_]*$/` since
> The <dollar-sign> shall anchor the expression (or optionally subexpression) to the end of the string being matched

drafts/25-xxx-specifications.txt
```diff
@@ -296,25 +296,31 @@ di07. The non-directive directive
 7 Predefined macros
 ===================
 
-pm01. Any macro name predefined by the implementation shall begin with a
+pm00. Macro names beginning with a leading underscore followed by an uppercase
+      letter or a second underscore are reserved to the processor. Reserved
+      names shall not be the subject of a #define or #undef preprocessing
+      directive within a program unit.
+      [Note: Any valid macro name matching the regular expression `/^_[A-Z][A-Za-z0-9_]*/`
+      is considered reserved.]
```
bonachea: The updated regex is incorrect because it fails to match names like `__reserved` that are reserved by the prose.
(I blame @gklimowicz ;-) )

```suggestion
      [Note: Any valid macro name matching the regular expression `/^_[A-Z_][A-Za-z0-9_]*/`
      is considered reserved.]
```

FWIW I'm ambivalent about the leading `^`, as I expect that will be understood. I'm not even sure this "note" is really necessary/beneficial, because it just duplicates what the rule prose already specifies.

bonachea (changes requested):
one new comment thread needs resolved

kc9jud:
> @kc9jud ping again, please stop force-pushing to rewrite reviewed commits.

oops, totally missed this comment before. Sorry about that, will stop -- different shops with different rules!

drafts/25-xxx-specifications.txt
kc9jud: Oops, you're right again.

Happy to remove the note, though IIUC notes are mostly just there to help the implementers but aren't part of the standard? It might just be that I'm weird and find it easier to parse a regex than a paragraph...

bonachea (approved):
Thanks for the fix! LGTM

drafts/25-xxx-specifications.txt
gklimowicz: I'm good with both `^` and `$` being used in the regex.

gklimowicz (approved):

kc9jud:
@bonachea @gklimowicz This is controversial enough that I'm not going to merge until after we discuss on Monday if this is a good idea in the first place.

PR #56, opened 2025-04-15
Initial draft of (most of) section 3, the #-directives

gklimowicz:
Missing is the longer section on the conditional directives (#if and family).

I haven't made a complete switch to avoiding the term "preprocessor" in favor of "preprocessing".

### DIFF
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -165,6 +165,8 @@ The preprocessor decomposes the source file into preprocessing tokens (see C 202
 Translation phases).
 As such, there is a specific lexicon of tokens recognized by
 the preprocessor (including unrecognizable tokens).
+As in C 2023, these tokens are recognized after any line and
+comment handling specified in the "Line" and "Comments" sections above.
 
 We use illustrative syntax to describe the directive specifications,
 and the translation behavior of the preprocess on Fortran comment lines
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -451,7 +451,7 @@ ew04. The processor produces a diagnostic message that includes
       the token-list.
 
 
-ew10. Additional behavior is processor-specific.
+ew10. Additional behavior is processor-dependent.
 
 
 3.7 The '#line' directive
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -507,7 +507,7 @@ pr20. The token-list is expanded.
 pr22. The #pragma directive causes the processor to behave
       in a processor-defined manner.
 
-pr24. Any #pragma that is not recognized by the implementation
+pr24. Any #pragma that is not recognized by the processor
       is ignored.
 
 
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -375,6 +375,7 @@ ud24.
 Example syntax:
       #include CHARACTER_STRING EOL
       #include <character-list> EOL
+      #include token-list EOL
 
 
 3.4.1 Static semantics specifications
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -409,7 +409,9 @@ in24. If the file is located, the processor replaces the #include
 
 in26. A #include directive may appear in an included file, up to a
       processor-defined nesting limit.
-
+in28. Unlike INCLUDE lines (see the section on "INCLUDE lines"), 
+      the #include directive is not prohibited from including the
+      same source file at a deeper level of nesting.
 
 3.5 The '#if', '#ifdef', '#ifndef', '#elif', '#elifdef',
       '#elifndef', '#else', '#endif' directives
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -384,7 +384,7 @@ Example syntax:
 in02. In the first form, the CHARACTER_STRING may be delimited with
       single-quote "'" or double-quote '"' characters.
 
-in04. A "_kind" specifier shall not appear following the CHARACTER_STRING.
+in04. CHARACTER_STRING shall not include a kind type parameter.
 
 in06. In the second form, the character-list is any sequence of
       processor-specific characters except EOL and `>`.
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -485,9 +485,6 @@ li24. In the second form, the #line directive causes the processor
       to change the presumed file name of the source file to be the
       contents of the CHARACTER_STRING.
 
-li25. In the second form, the CHARACTER_STRING must be delimited with
-      double-quote '"' characters (not single-quote "'" characters).
-
 3.8 The '#pragma' directive
 ---------------------------
 
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -440,7 +440,7 @@ error conditions in the input.
 
 ew02. The token-list is expanded.
 
-ew04. The preprocessor produces a diagnostic message that includes
+ew04. The processor produces a diagnostic message that includes
       the result of the token expansion.
 
 ew06. The #error directive causes the compilation process to end after
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -238,7 +238,7 @@ di08. The '#pragma' directive
 
 di09. The null directive
 
-di10. The non-directive directive
+di10. The processor-dependent directive
 
 
 
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -199,7 +199,7 @@ to11. The following tokens also appear in the illustrative syntax below.
       | ID                  | Regex [A-Za-z][A-Za-z0-9_]*               |
       | WHOLE_NUMBER        | Regex [0-9]+                              |
       | REAL_NUMBER         | Fortran 2023 R714 real-literal-constant.  |
-      | CHARACTER_STRING    | RegEx "[^"\n]*"                           |
+      | CHARACTER_STRING    | RegEx "[^"\n]*" where '\n' represents new-line. |
       | UNRECOGNIZED_TOKEN  | Any character not recognized above.       |
       | EOL                 | The new-line ending a directive line,     |
       |                     | after continuation processing (see li15). |
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -315,9 +315,9 @@ df04. The identifier-list is a comma-separated list of ID tokens.
 
 df06. No identifier may appear more than once in the identifier-list.
 
-df08. The identifier names in the identifier-list have "scope" in the
-      replacement list. (See the section "Macro recognition and expansion" for the
-      semantics of function-like macro expansion.)
+df08. The identifier names in the identifier-list define macro "parameters" that affect
+      macro expansion of the replacement list. (See the section "Macro recognition and 
+      expansion" for the semantics of function-like macro expansion.)
 
 df10. The replacement-list may be the empty sequence of tokens.
 
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -541,7 +541,7 @@ Example syntax:
 
 nd02. The token-list in a non-directive directive may not begin
       with any of the directive names appearing in the syntax above
-      ('define', 'undef', 'include,, etc.).
+      ('define', 'undef', 'include', etc.).
 
 
 3.10.2 Evaluation semantics specifications
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -388,9 +388,17 @@ in04. CHARACTER_STRING shall not include a kind type parameter.
 in06. In the second form, the character-list is any sequence of
       processor-specific characters except EOL and `>`.
 
+in08. In the third form, the token-list does not match the previous
+      two forms.
+
 
 3.4.2 Evaluation semantics specifications
 -----------------------------------------
+
+in18. In the third form, the token-list is processed as in normal
+      text (see the section on Macro Recognition and Expansion below).
+      The directive resulting after all replacements shall match one
+      of the previous two forms, and evaluation proceeds as such.
 
 in20. The preprocessor searches in processor-defined places for the
       file denoted by the CHARACTER_STRING or the character-list.
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -450,11 +450,6 @@ ew02. The token-list is not expanded.
 ew04. The processor produces a diagnostic message that includes
       the token-list.
 
-ew06. The #error directive causes the compilation process to end after
-      preprocessing the input files.
-
-ew08. The #warning directive allows compilation to continue after
-      preprocessing.
 
 ew10. Additional behavior is processor-specific.
 
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -384,7 +384,7 @@ Example syntax:
 --------------------------------------
 
 in06. In the second form, the character-list is any sequence of
-      processor-specific characters except EOL and `>`.
+      processor-dependent characters except EOL and `>`.
 
 in08. In the third form, the token-list does not match the previous
       two forms.
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -316,7 +316,7 @@ df04. The identifier-list is a comma-separated list of ID tokens.
 df06. No identifier may appear more than once in the identifier-list.
 
 df08. The identifier names in the identifier-list have "scope" in the
-      replacement list. (See the section on macro expansion for the
+      replacement list. (See the section "Macro recognition and expansion" for the
       semantics of function-like macro expansion.)
 
 df10. The replacement-list may be the empty sequence of tokens.
@@ -326,7 +326,7 @@ df12. Whitespace before or after the replacement-list is not
 
 df16. The "..." between the parentheses specifies that the function-like
       macro may be invoked with a variable number of arguments. (See the
-      section on macro expansion for the semantics of function-like macros
+      section "Macro recognition and expansion" for the semantics of function-like macros
       with a variable number of arguments.)
 
 
@@ -396,8 +396,8 @@ in08. In the third form, the token-list does not match the previous
 -----------------------------------------
 
 in18. In the third form, the token-list is processed as in Fortran
-      source fragments (see the section on Macro Recognition and
-      Expansion below). The directive resulting after all replacements
+      source fragments (see the section on "Macro recognition and
+      expansion" below). The directive resulting after all replacements
       shall match one of the previous two forms, and evaluation proceeds
       as such.
 
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -386,7 +386,7 @@ in02. In the first form, the CHARACTER_STRING may be delimited with
 in04. A "_kind" specifier shall not appear following the CHARACTER_STRING.
 
 in06. In the second form, the character-list is any sequence of
-      processor-specific characters.
+      processor-specific characters except EOL and `>`.
 
 
 3.4.2 Evaluation semantics specifications
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -443,7 +443,7 @@ ew02. The token-list is expanded.
 ew04. The preprocessor produces a diagnostic message that includes
       the result of the token expansion.
 
-ew06. The #error directive causes the compilation process to and after
+ew06. The #error directive causes the compilation process to end after
       preprocessing the input files.
 
 ew08. The #warning directive allows compilation to continue after
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -197,8 +197,10 @@ to11. The following tokens also appear in the illustrative syntax below.
       | ID                  | Regex [A-Za-z][A-Za-z0-9_]*               |
       | WHOLE_NUMBER        | Regex [0-9]+                              |
       | REAL_NUMBER         | Fortran 2023 R714 real-literal-constant.  |
-      | CHARACTER_STRING    | Fortran 2023 R724 char-literal-constant.  |
+      | CHARACTER_STRING    | RegEx "[^"\n]*"                           |
       | UNRECOGNIZED_TOKEN  | Any character not recognized above.       |
+      | EOL                 | The new-line ending a directive line,     |
+      |                     | after continuation processing (see li15). |
       |---------------------|-------------------------------------------|
 
 
@@ -381,9 +383,6 @@ Example syntax:
 3.4.1 Static semantics specifications
 --------------------------------------
 
-in02. In the first form, the CHARACTER_STRING may be delimited with
-      single-quote "'" or double-quote '"' characters.
-
 in04. CHARACTER_STRING shall not include a kind type parameter.
 
 in06. In the second form, the character-list is any sequence of
@@ -439,10 +438,10 @@ error conditions in the input.
 3.6.2 Evaluation semantics specifications
 -----------------------------------------
 
-ew02. The token-list is expanded.
+ew02. The token-list is not expanded.
 
 ew04. The processor produces a diagnostic message that includes
-      the result of the token expansion.
+      the token-list.
 
 ew06. The #error directive causes the compilation process to end after
       preprocessing the input files.
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -539,7 +539,7 @@ Example syntax:
 3.10.1 Static semantics specifications
 --------------------------------------
 
-nd02. The token-list in a non-directive directive may not begin
+nd02. The token-list in a processor-dependent directive may not begin
       with any of the directive names appearing in the syntax above
       ('define', 'undef', 'include', etc.).
 
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -334,7 +334,7 @@ df16. The "..." between the parentheses specifies that the function-like
 -----------------------------------------
 
 df30. This #define directive defines a function-like macro with parameters
-      (dummy arguments) as specified in the identifier-list and function
+      as specified in the identifier-list and function
       body as specified in the replacement list.
 
 df32. This #define directive does not scan the replacement-list
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -371,8 +371,6 @@ ud20. If no definition exists for the identifier ID, this directive has
 ud22. The definition for the object-like macro or function-like macro
       named by ID is removed.
 
-ud24.
-
 3.4 The '#include' directive
 ----------------------------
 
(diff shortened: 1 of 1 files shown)


kc9jud (changes requested):
Looks pretty good, I have a few corrections and a few questions.

kc9jud:
@gklimowicz did you mean to request reviews from the others? It looks like you just "assigned" it to us rather than requesting reviews. (I still don't understand what it means to "assign" a PR to someone...)

drafts/25-xxx-specifications.txt
```diff
@@ -206,64 +206,348 @@ to11. The following tokens also appear in the illustrative syntax below.
...
+
+do22. There is one namespace for all macro identifiers.
+
+do24. The #define directive does not scan the replacement-list
+      for macros to expand.
+
+do26. An identifier currently defined as an object-like macro shall not
+      be redefined by another #define preprocessing directive unless
+      the second definition is an object-like macro definition and
+      the two replacement lists are identical (as in C 2023).
```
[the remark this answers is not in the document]
gklimowicz: I believe it is an error. From C 2023, 6.10.4

> "Two replacement lists are identical if and only if the preprocessing tokens in both have the same
> number, ordering, spelling, and white-space separation, where all white-space separations are
> considered identical.
> An identifier currently defined as an object-like macro shall not be redefined by another #define
> preprocessing directive unless the second definition is an object-like macro definition and the two
> replacement lists are identical."

Now in the Fortran standard, "shall not" typically means "must not". But I have a dollar that says that some current implementations default to a warning.

drafts/25-xxx-specifications.txt
```diff
@@ -206,64 +206,348 @@ to11. The following tokens also appear in the illustrative syntax below.
...
+      #define ID(identifier-list, ...) replacement-list EOL
+
+
+3.2.1 Static semantics specifications
+--------------------------------------
+
+df02. The identifier ID must be immediately followed by the
+      left parenthesis '(' with no intervening whitespace.
+
+df04. The identifier-list is a comma-separated list of ID tokens.
```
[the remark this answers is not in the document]
gklimowicz: I agree. df04 doesn't specify anything about whitespace, so the usual whitespace rules apply. (Maybe we need to say more about that.)

df02 is the only place where the usual whitespace rules don't apply. That's the intention, at least.

drafts/25-xxx-specifications.txt
```diff
@@ -206,64 +206,348 @@ to11. The following tokens also appear in the illustrative syntax below.
...
+      named by ID is removed.
+
+ud24.
+
 3.4 The '#include' directive
 ----------------------------
 
+Example syntax:
+      #include CHARACTER_STRING EOL
+      #include <character-list> EOL
```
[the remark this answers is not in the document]
gklimowicz: At the discussion at a recent (the last?) meeting there was feedback that some didn't want the third option at all, and there were negative comments about even the pointy-bracket `<` ... `>` option.

gklimowicz:
Assigning was a mistake. Apparently it's easy for me to make this mistake repeatedly.

drafts/25-xxx-specifications.txt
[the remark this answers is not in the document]
kc9jud: I think having the token-list option is more important to me than having two different spellings of regular `#include`...

drafts/25-xxx-specifications.txt
```diff
@@ -206,64 +206,348 @@ to11. The following tokens also appear in the illustrative syntax below.
 3 #-Directives
```
bonachea, line 206: Please add `EOL` to the list of tokens appearing in the illustrative syntax. I suggest something like:

```
| EOL                 | The new-line ending a directive line,     |
|                     |  after continuation processing (see li15) |

File: drafts/25-xxx-specifications.txt (hunk 1 above)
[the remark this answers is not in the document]
Comment (bonachea) [reply drafts/25-xxx-specifications.txt:281]: I'm fine with this wording, implying it's an error.

Some implementations are permissive, but IMO deliberately changing a define without `#undef` is a terrible programming practice and should always be a hard error. If you're _intentionally_ changing a definition, you should always start with `#undef`. If you're not, then an error should alert you to the conflict.

File: drafts/25-xxx-specifications.txt [hunk 5]
```diff
@@ -206,64 +206,348 @@ to11. The following tokens also appear in the illustrative syntax below.
...
+
+ud20. If no definition exists for the identifier ID, this directive has
+      no effect.
+
+ud22. The definition for the object-like macro or function-like macro
+      named by ID is removed.
+
+ud24.
+
 3.4 The '#include' directive
```
bonachea, line 372:
```suggestion
3.4 The '#include' directive
```

drafts/25-xxx-specifications.txt
```diff
@@ -206,64 +206,348 @@ to11. The following tokens also appear in the illustrative syntax below.
...
+Example syntax:
+      #include CHARACTER_STRING EOL
+      #include <character-list> EOL
+
+
+3.4.1 Static semantics specifications
+--------------------------------------
+
+in02. In the first form, the CHARACTER_STRING may be delimited with
+      single-quote "'" or double-quote '"' characters.
```
bonachea, line 384: C does NOT allow single-quote to delimit the CHARACTER_STRING in `#include`, and both gcc and gfortran currently consider this a hard error.

I suspect this is an attempt to be more "Fortran-friendly" in matching Fortran character literal constant syntax, but we're not accepting other Fortran-specific character literal syntax (e.g. kind type parameters). However because C compilers (and some current Fortran compilers) will reject this syntax, I think allowing single-quote strings in `#include` encourages a non-portable programming practice.

Do we really want this extension?

```suggestion
in02. In the first form, the CHARACTER_STRING must be delimited with
      double-quote '"' characters (not single-quote "'" characters).
```

drafts/25-xxx-specifications.txt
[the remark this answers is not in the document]
bonachea: I also think the token-list option is important. It notably generalizes producer-consumer include patterns where one module produces a header file (with a module specific-name) that is consumed by a second module via include. This syntax specifically allows the consumer to avoid hard-coding the name of the shared header, allowing the producer to determine an appropriate file name.

Here's an self-contained example distilled from real code in one of my projects:

consumer.c:
```
#ifndef STRINGIFY
#define STRINGIFY_HELPER(x) #x
#define STRINGIFY(x) STRINGIFY_HELPER(x)
#endif

#ifdef PRODUCER_HEADER
/* use the token-list syntax of #include: */
#include STRINGIFY(PRODUCER_HEADER)
#endif

/* body of code that depends on #defines optionally provided in PRODUCER_HEADER */
```

build command, invoked by the producer to build its dependency consumer.o:
```
$ cc -DPRODUCER_HEADER=producer_config.h consumer.c
```
Where `producer_config.h` is a long, auto-generated header produced by a configure script in the producer.

drafts/25-xxx-specifications.txt
```diff
@@ -206,64 +206,348 @@ to11. The following tokens also appear in the illustrative syntax below.
...
+3.6.2 Evaluation semantics specifications
+-----------------------------------------
 
-3.5 The '#line' directive
+ew02. The token-list is expanded.
+
+ew04. The preprocessor produces a diagnostic message that includes
+      the result of the token expansion.
+
+ew06. The #error directive causes the compilation process to and after
```
bonachea, line 446:
```suggestion
ew06. The #error directive causes the compilation process to end after
```

drafts/25-xxx-specifications.txt
```diff
@@ -206,64 +206,348 @@ to11. The following tokens also appear in the illustrative syntax below.
...
+li04. The WHOLE_NUMBER shall not exceed 2147483647.
+
+
+3.7.2 Evaluation semantics specifications
+-----------------------------------------
+
+li20. In the first example, the #line directive causes the preprocessor
+      to behave as if the following sequence of Fortran source lines or
+      comment lines begins with a source line that has a line number
+      specified by the WHOLE_NUMBER.
```
bonachea, line 477: This wording seems to imply the altered numbering does not apply to directive lines, which it also does.

We've also switched from calling the syntax "forms" to "examples". Use "forms"

Rewording for more precision:
```suggestion
li20. In both forms, the #line directive causes the processor
      to behave as if the following sequence of lines begins with a 
      presumed line number specified by the WHOLE_NUMBER.
```

drafts/25-xxx-specifications.txt
```diff
@@ -206,64 +206,348 @@ to11. The following tokens also appear in the illustrative syntax below.
...
 
-3.6 The '#pragma' directive
+li22. In the second example, the #line directive causes the preprocessor
+      to behave as if the following sequence of Fortran source lines or
+      comment lines begins with a source line that has a line number
+      specified by the WHOLE_NUMBER.
+
```
bonachea, line 483:
```suggestion
```

drafts/25-xxx-specifications.txt
```diff
@@ -206,64 +206,348 @@ to11. The following tokens also appear in the illustrative syntax below.
...
+li24. In the second example, the #line directive causes the preprocessor
+      to change the presumed file name of the source file to be the
+      contents of the CHARACTER_STRING.
+
```
bonachea, line 487:
```suggestion
li24. In the second form, the #line directive causes the processor
      to change the presumed file name of the source file to be the
      contents of the CHARACTER_STRING.

li25. In the second form, the CHARACTER_STRING must be delimited with
      double-quote '"' characters (not single-quote "'" characters).
```

bonachea (changes requested):
Thanks for the great start on this @gklimowicz !

I've added a number of requests for improvement.

drafts/25-xxx-specifications.txt
```diff
@@ -206,64 +206,348 @@ to11. The following tokens also appear in the illustrative syntax below.
...
+      #include <character-list> EOL
+
+
+3.4.1 Static semantics specifications
+--------------------------------------
+
+in02. In the first form, the CHARACTER_STRING may be delimited with
+      single-quote "'" or double-quote '"' characters.
+
+in04. A "_kind" specifier shall not appear following the CHARACTER_STRING.
```
bonachea, line 386: According to F23, the optional kind type parameter actually comes _before_ a character literal constant. However the positioning is also an irrelevant detail.

```suggestion
in04. CHARACTER_STRING shall not include a kind type parameter.
```

However at this point it might just be easier to simply redefine `CHARACTER_STRING`:

```
 | CHARACTER_STRING    | RegEx "[^"\n]*"  |
```
which allows us to delete this rule and the one above it.

bonachea (changes requested):
A few more...

drafts/25-xxx-specifications.txt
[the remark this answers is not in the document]
bonachea: > having the token-list option is more important to me than having two different spellings of regular `#include`

Upon further reflection, I think the Fortran `INCLUDE` line _may already_ give us access to the desired functionality. In particular, a Fortran `INCLUDE` line is a regular (non-directive) Fortran source fragment, and thus (unless we say otherwise) is also subject to regular macro expansion.

As such, I think my prior example could have been written for FPP as:
```
#ifndef STRINGIFY
#define STRINGIFY_HELPER(x) #x
#define STRINGIFY(x) STRINGIFY_HELPER(x)
#endif

#ifdef PRODUCER_HEADER
! use a Fortran INCLUDE line:
INCLUDE STRINGIFY(PRODUCER_HEADER)
#endif

! body of code that depends on #defines optionally provided in PRODUCER_HEADER 
```

If we adequately specify the above pattern must work, then I suspect that spelling can (with mostly trivial differences) provide the missing capability for Fortran codes. With that in-place I believe the (controversial?) `# include pp-tokens new-line` form becomes mostly redundant and can be safely omitted.

I think the key to making this work is a new subsection somewhere about how Fortran `INCLUDE` lines are handled. We probably need this anyhow because of the interaction with preprocessing in general, and the macro namespace in particular. In that section we add a rule saying that `INCLUDE` lines are subject to macro expansion (just like any other non-directive line), and that the argument is interpreted after macro expansion.

Thoughts @kc9jud ?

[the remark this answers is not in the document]
gklimowicz: You're right that we're missing the section about `INCLUDE`. I believe that when we add it that we allow the same syntax for `INCLUDE` and `#include`, but I haven't yet looked closely at any subtleties.

drafts/25-xxx-specifications.txt
```diff
@@ -206,64 +206,348 @@ to11. The following tokens also appear in the illustrative syntax below.
...
+
+
+3.6.2 Evaluation semantics specifications
+-----------------------------------------
 
-3.5 The '#line' directive
+ew02. The token-list is expanded.
+
+ew04. The preprocessor produces a diagnostic message that includes
+      the result of the token expansion.
```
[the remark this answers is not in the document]
gklimowicz: Well that sure makes sense.

drafts/25-xxx-specifications.txt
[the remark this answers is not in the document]
kc9jud: Got it -- I was slightly concerned because I didn't know if "comma-separated" allowed for spaces around commas. (In the detailed spec I suppose we will have to define "comma-separated" more precisely.)

drafts/25-xxx-specifications.txt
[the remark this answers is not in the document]
kc9jud: @bonachea That's fine by me, though at that point I'm not sure why we would allow `INCLUDE EXPAND_ME` but not `#include EXPAND_ME`...

drafts/25-xxx-specifications.txt
bonachea, line 1: In today's meeting we resolved to add a new placeholder section for the Fortran `INCLUDE` line, possibly in a separate PR.

I suggest we iterate on the actual content for that new section in a follow-on PR.

drafts/25-xxx-specifications.txt
bonachea: In today's meeting we resolved to define the `CHARACTER_STRING` token as suggested above, and adjust all references accordingly.

bonachea (changes requested):
Recording resolutions from today's Zoom meeting

drafts/25-xxx-specifications.txt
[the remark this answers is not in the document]
gklimowicz: I've written up issue #58 to track this for another pull request.

drafts/25-xxx-specifications.txt
gklimowicz: Agreed. See #58.

drafts/25-xxx-specifications.txt
```diff
@@ -197,73 +197,361 @@ to11. The following tokens also appear in the illustrative syntax below.
       | ID                  | Regex [A-Za-z][A-Za-z0-9_]*               |
       | WHOLE_NUMBER        | Regex [0-9]+                              |
       | REAL_NUMBER         | Fortran 2023 R714 real-literal-constant.  |
-      | CHARACTER_STRING    | Fortran 2023 R724 char-literal-constant.  |
+      | CHARACTER_STRING    | RegEx "[^"\n]*"                           |
```
bonachea, line 200: Following up from a discussion point in our meeting, there was a question about corner-case inputs like this:

```
#include "stddef.h\
"
```

It appears this *IS* permitted by C23, because line conjoining removes the backslash-newline _before_ tokenization of the character string. I think we are specifying the same behavior, meaning the regexs in this token list (eg. `CHARACTER_STRING`) are applied AFTER line conjoining.

Agreed?

drafts/25-xxx-specifications.txt
```diff
@@ -197,73 +197,361 @@ to11. The following tokens also appear in the illustrative syntax below.
...
+
+in08. In the third form, the token-list does not match the previous
+      two forms.
+
+
+3.4.2 Evaluation semantics specifications
+-----------------------------------------
+
+in18. In the third form, the token-list is processed as in normal
+      text (see the section on Macro Recognition and Expansion below).
```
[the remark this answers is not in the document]
gklimowicz: Will make this change and push a new set of commits.

drafts/25-xxx-specifications.txt
```diff
@@ -197,73 +197,361 @@ to11. The following tokens also appear in the illustrative syntax below.
...
+      semantics of function-like macro expansion.)
+
+df10. The replacement-list may be the empty sequence of tokens.
+
+df12. Whitespace before or after the replacement-list is not
+      considered to be part of the replacement-list.
+
+df16. The "..." between the parentheses specifies that the function-like
+      macro may be invoked with a variable number of arguments. (See the
+      section on macro expansion for the semantics of function-like macros
```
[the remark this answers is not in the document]
gklimowicz: I tend to avoid specific subsection numbers in systems that don't support reference links (such as this kind of .txt paper). Maybe a better compromise might be to use the full section name:
```suggestion
      macro may be invoked with a variable number of arguments. (See the
      "Macro recognition and expansion" section below for the semantics of function-like macros
```

drafts/25-xxx-specifications.txt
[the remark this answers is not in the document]
gklimowicz: Hmmm. If we add Dan's suggestion above about token recognition being performed after conjoining (which I think is the correct stance), then the `CHARACTER_STRING` regex doesn't have to mention new-line at all. EOL is a token that the lexer produces at the end of a directive line. So the raw text line
```C
# define X "I'm starting a string
continued on another line"
```
does not contain a well-formed `CHARACTER_STRING` token. It is rejected in lexical analysis as ill-formed. It's just something else.

(You could make the case (I won't) that the tokens represented above are `#`, `DEFINE`, `ID("X")`, `DOUBLE_QUOTE` `ID("I")`, `SINGLE_QUOTE`, `ID("m")`, `ID("starting")`, `ID("a")`, `ID("string")` `EOL`, `ID("continued")`, `ID("on")`, `ID("another")`, `ID("line")` `EOL`)

But the raw text below, before conjoining, does contain a valid `CHARACTER_STRING` token after conjoining:
```C
# define X "I'm starting a string\
"
```

gklimowicz: I think that is right. Conjoining takes precedence over tokenization. I thing this also handles Patrick's concern about `CHARACTER_STRING` regex below (and then some).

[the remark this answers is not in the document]
gklimowicz: By the way, we don't explicitly use `\n` anywhere else, but we do use the C standard's term _new-line_ where (like them) we want to avoid concerning ourselves with line termination encodings. I just couldn't put _new-line_ in the original regex.

drafts/25-xxx-specifications.txt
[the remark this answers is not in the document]
gklimowicz: Addressed in the updated commits. I made a couple other changes to make the use consistent throughout the document.

drafts/25-xxx-specifications.txt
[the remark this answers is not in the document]
gklimowicz: I could add a note to that effect.
```suggestion
      | CHARACTER_STRING    | RegEx "[^"\n]*" where \n represents new-line |
```

drafts/25-xxx-specifications.txt
```diff
@@ -197,73 +199,357 @@ to11. The following tokens also appear in the illustrative syntax below.
...
 
-3.3 The '#if', '#ifdef', '#ifndef', '#elif', '#elifdef',
+in22. It is an error if the processor cannot locate the specified file.
+
+in24. If the file is located, the processor replaces the #include
+      directive line with the contents of that file.
+
+in26. A #include directive may appear in an included file, up to a
+      processor-defined nesting limit.
+
```
bonachea, line 414: One subtle but important way in which `#include` differs from `INCLUDE` lines, the latter specifies:

> Inclusion of the source text referenced by an INCLUDE line shall not, at any level of nesting, result in inclusion of the same source text.

Given that Fortran precedent, I think that (at least for the purposes of this document) it's worth explicitly calling out the fact that `#include` does NOT have that restriction. Eg:

```suggestion
in28. Unlike INCLUDE lines (see the section on "INCLUDE lines"), 
      the #include directive is not prohibited from including the
      same source file at a deeper level of nesting.
```

drafts/25-xxx-specifications.txt
```diff
@@ -197,73 +199,357 @@ to11. The following tokens also appear in the illustrative syntax below.
...
+ew02. The token-list is not expanded.
+
+ew04. The processor produces a diagnostic message that includes
+      the token-list.
+
+ew06. The #error directive causes the compilation process to end after
+      preprocessing the input files.
+
+ew08. The #warning directive allows compilation to continue after
+      preprocessing.
```
bonachea, line 459: Dictating exactly what happens to "the compilation process" here may be over-reach.

C23 simply says:
> causes the implementation to produce a diagnostic message that includes the specified sequence of preprocessing tokens.

I think implementations may even differ in whether they stop immediately upon encountering the first `#error`, or whether they try to continue further. Definitely the text above implying implementations are required to _complete_ preprocessing of all inputs after encountering an `#error` seems wrong.

What if anything does Fortran say about how processors are required to handle an compilation error after detection?
4.2 includes text like:

> A Fortran processor shall: ...
contain the capability to detect and report the use within a submitted program unit of a form or relationship that is not permitted by the numbered syntax rules or constraints,

I suspect we just want to delete these two rules, leaving the "processor-specific" rule below.



```suggestion
```

drafts/25-xxx-specifications.txt
```diff
@@ -197,73 +199,357 @@ to11. The following tokens also appear in the illustrative syntax below.
...
+
+ew10. Additional behavior is processor-specific.
```
bonachea, line 461: "processor-specific" does not appear to be proper "Fortran-ese".

```suggestion
ew10. Additional behavior is processor-dependent.
```

drafts/25-xxx-specifications.txt
```diff
@@ -197,73 +199,357 @@ to11. The following tokens also appear in the illustrative syntax below.
...
+      #include CHARACTER_STRING EOL
+      #include <character-list> EOL
+      #include token-list EOL
+
+
+3.4.1 Static semantics specifications
+--------------------------------------
+
+in06. In the second form, the character-list is any sequence of
+      processor-specific characters except EOL and `>`.
```
bonachea, line 389:
```suggestion
      processor-dependent characters except EOL and `>`.
```

bonachea (changes requested):
Coming along nicely, noted a few more areas for improvement

drafts/25-xxx-specifications.txt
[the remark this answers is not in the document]
gklimowicz: I added the note in the updated commit. (I didn't edit the table row length; we'll fix that up in final edits.)

drafts/25-xxx-specifications.txt
```diff
@@ -197,73 +199,357 @@ to11. The following tokens also appear in the illustrative syntax below.
...
+      section "Macro recognition and expansion" for the semantics of function-like macros
+      with a variable number of arguments.)
+
+
+3.2.2 Evaluation semantics specifications
+-----------------------------------------
+
+df30. This #define directive defines a function-like macro with parameters
+      (dummy arguments) as specified in the identifier-list and function
+      body as specified in the replacement list.
```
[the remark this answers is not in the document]
gklimowicz: I won't pretend to like using "dummy argument" here in the first place. So I am good with this change.

drafts/25-xxx-specifications.txt
kc9jud: @bonachea does that mean that Fortran `INCLUDE` has the equivalent of `#pragma once` in every included file? In C one always has `#ifdef` guards or `#pragma once` but this sounds like Fortran doesn't need that?

bonachea: @gklimowicz asks:
> Do we have a use case for this? Otherwise, my naive preference is to behave as Fortran `INCLUDE` does.

Yes absolutely, this detail is **_very_** important in the context of CPP `#include`.

 In well-written C/C++ headers the standard idiom is every header has the following structure (after comments are removed).
my_header.h:
```
#ifndef _MY_HEADER_H
#define _MY_HEADER_H

// .. entire body of the header...

#include <foo.h> /* for example.. */

#endif
```
With this idiom in-place, recursive inclusion of a header (possibly through one or more intermediaries) becomes harmless and well-defined (the nested expansion is empty).

This is a very important feature property because: with larger numbers of headers the dependency include graph encompassing them all can become very complicated, and information about what each header needs might be a local property that changes over time. With the CPP semantics of `#include` and idiom above, if `foo.h` ever expands to contain `#include <my_header.h>` (possibly through some nested intermediary header) this nested/repeated include is harmless and ignored.

@kc9jud asks:
> does that mean that Fortran INCLUDE has the equivalent of #pragma once in every included file? In C one always has #ifdef guards or #pragma once but this sounds like Fortran doesn't need that?

`#pragma once` is a NON-standard conforming mechanism provided by many C/C++ compilers to solve the same problem. The pattern I've shown above is a standards-conforming solution.

My understanding of the Fortran `INCLUDE` line specification (see above) is NOT equivalent to `#pragma once`. A nested Fortran `INCLUDE` line is a "shall not" constraint violation, and thus simply erroneous (which is NOT the behavior we want for `#include`).

sblionel: The Fortran standard says for INCLUDE:

32 5 The effect of the INCLUDE line is as if the referenced source text physically replaced the INCLUDE line prior
33 to program processing. Included text may contain any source text, including additional INCLUDE lines; such
34 nested INCLUDE lines are similarly replaced with the specified source text. The maximum depth of nesting of
35 any nested INCLUDE lines is processor dependent. **Inclusion of the source text referenced by an INCLUDE line
36 shall not, at any level of nesting, result in inclusion of the same source text.**

This is, of course, a restriction on what is standard-conforming source and is not a constraint, but rather normative text. Since we have already said that we want `#include` to behave like `INCLUDE`, I think it should follow the same restriction.

kc9jud: @bonachea is there actually a use case for allowing cyclic dependency graphs with either `INCLUDE` or `#include` though? I'll be honest that I've usually found the need for an include guard to be a shortcoming of CPP rather than a feature.

bonachea: FWIW both gfortran-14 and flang-20 handle this correctly in a manner matching the CPP behavior. Here's a toy example where there are two headers, and including either of them has the effect of including them both:

```
==> bar.h <==
#ifndef INCLUDED_BAR_H
#define INCLUDED_BAR_H

#include "foo.h"

integer :: bar

#endif

==> foo.h <==
#ifndef INCLUDED_FOO_H
#define INCLUDED_FOO_H

#include "bar.h"

integer :: foo

#endif

==> test.F90 <==
#include "foo.h"
```
```
$ gfortran -E test.F90
# 1 "test.F90"
# 1 "<built-in>"
# 1 "<command-line>"
# 1 "test.F90"

# 1 "foo.h" 1




# 1 "bar.h" 1




# 1 "foo.h" 1
# 5 "bar.h" 2

integer :: bar

# 5 "foo.h" 2

integer :: foo

# 2 "test.F90" 2
```
```
$ flang -E test.F90   
#line "./bar.h" 6
      integer :: bar
#line "./foo.h" 6
      integer :: foo

```

Thus existing practice also argues for keeping CPP's permissiveness on recursive `#include`.

drafts/25-xxx-specifications.txt
gklimowicz: I'm fine with this.

drafts/25-xxx-specifications.txt
gklimowicz: I should just put that in my auto-correct. :-)

drafts/25-xxx-specifications.txt
gklimowicz: I'm OK with this.

bonachea (approved):

kc9jud (approved):
[Make it so](https://www.youtube.com/watch?v=FaLyasJPyUU)

bonachea:
> [Make it snow](https://www.youtube.com/watch?v=xteKObnaA2c)

PR #57, opened 2025-04-15
Reserve `__STDF*` specifically to the standard

kc9jud:
This goes the next step beyond reserving names beginning with underscores to the processor, and additionally reserves macro names beginning with `__STDF` to the standard. It forbids processors from defining macros beginning with `__STDF` unless explicitly listed in the standard. With #55 these names are already forbidden in program code.

This should be merged only if #55 is merged.


bonachea (approved):
I agree with the new pm03 rule added here, and that this PR should be merged after #55

gklimowicz (approved):

PR #61, opened 2025-05-05
Rough draft of Section 3.5 conditionals: #if, #elif, etc.

gklimowicz:
Initial draft of Section 3.5. Added illustrative syntax, static semantics and evaluation semantics, up to but not including controlling expression evaluation. (That will be an another section.)

I tried to use the terminology from the C 2023 standard, but reviewers will surely find lapses and potential improvements.

### DIFF
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -494,8 +494,8 @@ ix12. In the descriptions below, <group> constructs may be "skipped".
         - Conditional directives within the <group> are recognized only
           to maintain proper nesting of conditional directives.
         - No nested directives in the <group> are processed.
-        - No Fortran comment lines or source lines in the <group>
-          are examined for macro expansion.
+        - No macro expansion takes place in directive lines, Fortran
+          comment lines or source lines in the <group>.
         - No skipped lines of any kind in the <group> are made available
           subsequent processing by the processor.
 
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -414,20 +414,132 @@ in28. Unlike INCLUDE lines (see the section on "INCLUDE lines"),
       the #include directive is not prohibited from including the
       same source file at a deeper level of nesting.
 
+
 3.5 The '#if', '#ifdef', '#ifndef', '#elif', '#elifdef',
-      '#elifndef', '#else', '#endif' directives
+      '#elifndef', '#else', '#endif' conditional directives
 ---------------------------------------------------------
 
-Example syntax:
+Example syntax (extra spacing for illustration purposes only):
+    General form:
+        <if-head>  <lines>                        #endif EOL
+        <if-head>  <lines>  <if-elif>             #endif EOL
+        <if-head>  <lines>  <if-elif>  <if-else>  #endif EOL
 
+    Where:
+        <if-head> is one of:
+            #if token-list EOL  <lines>
+            #ifdef ID EOL       <lines>
+            #ifndef ID EOL      <lines>
+
+        <if-elif> is zero or more of:
+            #elif     token-list EOL  <lines>
+            #elifdef  ID EOL          <lines>
+            #elifndef ID EOL          <lines>
+
+        <if-else> is zero or one of:
+            #else EOL  <lines>
+
+        <lines> are zero or more of:
+            directive lines
+            Fortran comment lines
+            Fortran source lines
 
 
 3.5.1 Static semantics specifications
---------------------------------------
+-------------------------------------
+
+if05. A #if, ##ifdef, or #ifndef directive signals the start
+      of a list of conditional directives.
+
+if10: Whitespace must immediately follow the directives tokens
+      'ifdef', 'ifndef', 'elifdef', and 'elifndef' to separate
+      them from the ID tokens.
+
+if15. The list of #elif, #elifdef, and #elifndef directives, and
+      the #else directive (if present) are part of the same list
+      of directives introduced by the #if directive.
+
+if20. The chain of conditional directives ends with the #endif
+      directive.
+
+if25. Within a conditional directive list, a #if, #ifdef, or
+      #endif directive introduces a new list of conditional
+      directives, at a new nesting level, within the enclosing
+      conditional directive list.
+
+if30. The conditional directive lists must properly nest.
+
+if35. As shown in the illustrative syntax, there should be
+      no unmatched #endif directives.
+
+if40. As shown in the illustrative syntax, there should be
+      no more than one #else directive in a list of conditional
+      directives at any nesting level.
 
 
 3.5.2 Evaluation semantics specifications
 -----------------------------------------
+
+
+ix05. The #ifdef and #ifndef macros are evaluated as if
+      they had been written "#if defined(ID)" and
+      "#if !defined(ID)" respectively.
+
+ix10. The #elifdef and #elifndef macros are evaluated as if
+      they had been written "#elif defined(ID)" and "#elif !defined(ID)"
+      respectively.
+
+ix15. The token-lists in #if and #elif are processed as described
+      in the section "Macro recognition and expansion".
+
+ix20. After expansion, the resulting token list in #if and #elif
+      directives must represent a valid integer expression as
+      described in the section "Expressions allowed in #if and
+      #elif directives".
+
+ix25. The resulting token list is parsed and evaluated according
+      to the semantic rules described in section "Expressions allowed
+      in #if and #elif directives".
+
+ix30. If the expression in a #if evaluates to a C true value, the
+      <lines> immediately following the #if participate in preprocessing.
+      They are subject to preprocessing expansion,  substitution, and
+      evaluation of processing directives in the <lines>.
+
+ix40. Further preprocessing when #if evaluates to C true proceeds.
+      Subsequent #elif and #else directives at the same level of
+      nesting are skipped, and do not participate in preprocessing
+      except to be examined for proper nesting of conditional directives.
+
+ix45. If the expression in a #if evaluates to a C false value (zero),
+      the lines immediately following the #if directive are skipped, and
+      examined only to determine nesting of conditionals. No token
+      expansion occurs. No Fortran comment lines or source lines are
+      made available to the processor. Processing continues with
+      processing any #elif directives at the same level of nesting, as
+      described below.
+
+ix50. When the #if expression evaluates to C false, the #elif directives
+      at the same nesting level are evaluated in the order they appear
+      in the source. Those whose expression evaluate to false are
+      skipped in the same manner as when a #if clause evaluates to false.
+
+ix55. When a #elif expression evaluates to C true, the <lines>
+      immediately following the #elif participate in preprocessing
+      just as they do when the expression in a #if evaluates to C true.
+      Subsequent #elif and #else directives at the same nesting
+      level are skipped, as described for #if.
+
+ix60. When no #if expression or #elif expression evaluates to C true
+      and a #else directive is present at the same nesting level,
+      the <lines> between the #else directive and the matching #endif
+      directive are subject to preprocessing, expansion and substitution
+      as they would be in #if true processing.
+
+ix65. When no #if expression or #elif expression evaluates to C true
+      and no #else directive is present at the same nesting level,
+      there are no <lines> in the list of conditional directives
+      subject to preprocessing.
 
 
 3.6 The '#error' and '#warning' directives
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -59,7 +59,7 @@ li17. Preprocessor directive continuation processing described by the prior rule
 li19. The maximum length of a preprocessor directive (including
       continuation text) is one million characters.
 
-li21. A source file that ends with a directive line shall neither end with a '\', 
+li21. A source file that ends with a directive line shall neither end with a '\',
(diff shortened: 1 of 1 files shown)


gklimowicz:
Apparently, Gary does not know how to do a squash commit.

gklimowicz:
From email sent to the team, which may be useful for reviewers:

This PR covers only the directives in the #if family.

As these directives are highly interdependent, I didn’t write a separate section on each directive. Instead, there is a single omnibus illustrative syntax that includes all the directives, and the order and quantity in which they can appear.

As with the other directives, there is a subsection on the “static semantics”, the rules that must be followed in the _form_ the directives take. It covers loosely the proper nesting of conditional directives.

There is also the customary subsection on “evaluation semantics”: what happens when we _act on_ the directives.

Nesting syntax and semantics get a bit complicated. There is some text about that, but it is not as formal as I would like. (I may have to wait until the syntax paper to spell it out in detail.)

drafts/25-xxx-specifications.txt
```diff
@@ -413,22 +413,141 @@ in28. Unlike INCLUDE lines (see the section on "INCLUDE lines"),
...
-      '#elifndef', '#else', '#endif' directives
+      '#elifndef', '#else', '#endif' conditional directives
 ---------------------------------------------------------
 
-Example syntax:
+Example syntax (extra spacing for illustration purposes only):
+    General form:
+        <if-head>  <lines>                        #endif EOL
+        <if-head>  <lines>  <if-elif>             #endif EOL
+        <if-head>  <lines>  <if-elif>  <if-else>  #endif EOL
```
bonachea, line 425: Duplicate `<lines>` in `<if-head>`
```suggestion
        <if-head>                        #endif EOL
        <if-head>  <if-elif>             #endif EOL
        <if-head>  <if-elif>  <if-else>  #endif EOL
```

drafts/25-xxx-specifications.txt
```diff
@@ -413,22 +413,141 @@ in28. Unlike INCLUDE lines (see the section on "INCLUDE lines"),
...
+            directive lines
+            Fortran comment lines
+            Fortran source lines
 
 
 3.5.1 Static semantics specifications
---------------------------------------
+-------------------------------------
+
+if05. A #if, ##ifdef, or #ifndef directive signals the start of a list
```
bonachea, line 450:
```suggestion
if05. A #if, #ifdef, or #ifndef directive signals the start of a list
```

drafts/25-xxx-specifications.txt
```diff
@@ -413,22 +413,141 @@ in28. Unlike INCLUDE lines (see the section on "INCLUDE lines"),
...
+      of conditional directives.
+
+if10: Whitespace must immediately follow the directives tokens
+      'ifdef', 'ifndef', 'elifdef', and 'elifndef' to separate them
+      from the ID tokens.
```
bonachea, line 455: This property is roughly true for (nearly?) every preprocessing directive, not even just those in this section. Stating it here for only three makes them seem like an exception. (I say roughly because what's actually required is a token boundary, which might or might not include whitespace)

I think what we actually _mean_ here is that standard directives are only recognized as such if the token immediately following the `#` introducing a directive line exactly matches one of the standard directive names. If it does not match for any reason (e.g. due to lack of whitespace) then the directive does not conform to one of the standard directives and is instead parsed as a processor-dependent directive.

We should probably directly state something to this effect in section 3.0 and then delete this rule.

drafts/25-xxx-specifications.txt
```diff
@@ -413,22 +413,141 @@ in28. Unlike INCLUDE lines (see the section on "INCLUDE lines"),
...
+if40. As shown in the illustrative syntax, there should be no more
+      than one #else directive in a list of conditional directives at
+      any nesting level.
 
 
 3.5.2 Evaluation semantics specifications
 -----------------------------------------
 
 
+ix05. The #ifdef and #ifndef macros are evaluated as if they had been
```
bonachea, line 482:
```suggestion
ix05. The #ifdef and #ifndef directives are evaluated as if they had been
```

drafts/25-xxx-specifications.txt
```diff
@@ -413,22 +413,141 @@ in28. Unlike INCLUDE lines (see the section on "INCLUDE lines"),
...
+      directive.
+
+if25. Within a conditional directive list, a #if, #ifdef, or #endif
+      directive introduces a new list of conditional directives, at a new
+      nesting level, within the enclosing conditional directive list.
+
+if30. The conditional directive lists must properly nest.
+
+if35. As shown in the illustrative syntax, there should be no
+      unmatched #endif directives.
```
bonachea, line 471: This property can be directly derived from the pseudo-grammar above, so repeating it here is redundant. Worse, we only state one such property but omit others, for example: 

1. there should be no `<if-head>` without a matching `#endif`, 
2. there should be no `<if-elif>` without a matching `<if-head>` and `#endif`, 
3. there should be no `#else` without a matching `<if-head>` and `#endif`, 

which makes this one seem like an outlier. We should either list all such analogous grammatical properties or none.

drafts/25-xxx-specifications.txt
```diff
@@ -413,22 +413,141 @@ in28. Unlike INCLUDE lines (see the section on "INCLUDE lines"),
...
+      value, the <lines> immediately following the #if participate in
+      preprocessing. They are subject to preprocessing expansion,
+      substitution, and evaluation of processing directives in the
+      <lines>.
+
+ix40. Further preprocessing when #if controlling expression evaluates
+      to a nonzero value proceeds. Subsequent #elif and #else
+      directives at the same level of nesting are skipped, and do not
+      participate in preprocessing except to be examined for proper
+      nesting of conditional directives.
```
bonachea, line 514: I'm unhappy with the structure of how we describe the controlled `<lines>` here and in the subsequent rules. 

In particular, the definition of phrases like "<lines> immediately following the #if" and "<lines> immediately following the #elif" is too ambiguous, especially if those `<lines>` themselves contain directives (and possibly nested conditionals).

I think we should instead use the grammar non-terminals to help make this more precise. For example:

```suggestion
ix30. If the controlling expression in an <if-head> construct evaluates to a nonzero
      value, then the <lines> contained within that <if-head> construct participate in
      preprocessing. They are subject to preprocessing expansion,
      substitution, and evaluation of processing directives in those
      <lines>.

ix40. Further preprocessing when an <if-head> controlling expression evaluates
      to a nonzero value proceeds. Subsequent <if-elif> and <if-else>
      constructs at the same level of nesting are skipped, and the <lines>
      contained within those constructs do not participate in preprocessing 
      except to be examined for proper nesting of conditional directives.
```
and similarly for the next four rules.

Thoughts?

bonachea (changes requested):
My initial rough review of the rough draft ;-)

drafts/25-xxx-specifications.txt
LorriMenard: Ha, I saw that one too, and was afraid it was something from a meeting I missed.

drafts/25-xxx-specifications.txt
LorriMenard: Section 2.3 is titled "Significance of whitespace" - would this fit there?

drafts/25-xxx-specifications.txt
gklimowicz: Thanks. I missed this when refactoring the grammar on the fly.

drafts/25-xxx-specifications.txt
```diff
@@ -413,22 +413,141 @@ in28. Unlike INCLUDE lines (see the section on "INCLUDE lines"),
...
+      written "#if defined(ID)" and "#if !defined(ID)" respectively.
+      (For brevity, the descriptions of #ifdef and #ifndef directives
+      below are omitted, and assume this transformation.)
+
+ix10. The #elifdef and #elifndef macros are evaluated as if they had
+      been written "#elif defined(ID)" and "#elif !defined(ID)"
+      respectively.
+
+ix15. The token-lists in #if and #elif are processed as described in
+      the section "Macro recognition and expansion".
```
bonachea, line 492: As noted in the meeting, we also need to clarify whether macro expansion occurs within an `#elif` directive line if we already know that "group" is being skipped

drafts/25-xxx-specifications.txt
gklimowicz: I thought I replied that Section 2.3 seems like an excellent place for it. I will generate a pull request for section 2 for this.

drafts/25-xxx-specifications.txt
gklimowicz: I'm OK with letting the grammar speak for itself here, as long as we believe J3 as a whole won't feel like it's under-specified.

drafts/25-xxx-specifications.txt
gklimowicz: I like these updates to the phrasing. I am OK committing the phrasing for this one as suggested, and then updating the PR with new commits to add changes for the remaining rules.

drafts/25-xxx-specifications.txt
```diff
@@ -413,22 +413,141 @@ in28. Unlike INCLUDE lines (see the section on "INCLUDE lines"),
...
+      directives at the same level of nesting are skipped, and do not
+      participate in preprocessing except to be examined for proper
+      nesting of conditional directives.
+
+ix45. If the controlling expression in a #if evaluates to a zero
+      value, the lines immediately following the #if directive are
+      skipped, and examined only to determine nesting of conditionals.
+      Within these lines, no token expansion occurs, and no Fortran
+      comment lines nor source lines are made available for subsequent
+      processing by the processor. Preprocessing continues with any
```
kc9jud, line 521: Do we want to specifically call out that macros in the body are not evaluated in any way, in addition to the lack of expansion and source lines? I'm thinking of things like `#error`, `#warning`, and `#pragma` which aren't really about token expansion or source lines.

gklimowicz: You're right. We should be clear that in the skipped lines,

- Nested conditional directives are recognized but no evaluated
- No other directives are evaluated
- No macro expansion occurs
- No directive lines, Fortran source lines, or Fortran comment lines are made available for subsequent processing

drafts/25-xxx-specifications.txt
gklimowicz: Remove if35.

drafts/25-xxx-specifications.txt
gklimowicz: We will make these changes. And others, regarding use of `[ ]` for optional things, and _-list_ for lists.

drafts/25-xxx-specifications.txt
gklimowicz: Make sure that skipped parts act as if they aren't there, and have no effect on the preprocessing.

drafts/25-xxx-specifications.txt
gklimowicz: Added text in today's update. Will request further reviews.

drafts/25-xxx-specifications.txt
gklimowicz: Changes have been made for this in today's update. Will request further reviews.

drafts/25-xxx-specifications.txt
gklimowicz: Added text in today's update. This caused all the subsequent specifications on "skipping" and "participating in further preprocessing" to change. Will request further reviews.

drafts/25-xxx-specifications.txt
gklimowicz: Added text in today's update. Will request further reviews.

drafts/25-xxx-specifications.txt
gklimowicz: Made changes in today's update. Will request further reviews.

drafts/25-xxx-specifications.txt
```diff
@@ -232,14 +237,16 @@ di03. The '#undef' directive
...
+      '#elifndef', '#else', '#endif' conditional directives
 
 di06. The '#error' and '#warning' directives
 
 di07. The '#line' directive
 
 di08. The '#pragma' directive
 
+
+
```
kc9jud, line 249: Stray blank lines?

kc9jud (approved):
LGTM!

drafts/25-xxx-specifications.txt
gklimowicz: Fixed in new push.

bonachea (changes requested):
Apologies for the delayed review. Here's a few more nitpicks and then I think this PR is probably ready to merge.

drafts/25-xxx-specifications.txt
```diff
@@ -413,22 +418,146 @@ in28. Unlike INCLUDE lines (see the section on "INCLUDE lines"),
...
+if05. A #if, #ifdef, or #ifndef directive signals the start of a list
+      of conditional directives.
+
+if15. The list of #elif, #elifdef, and #elifndef directives, and the
+      #else directive (if present) are part of the same list of
+      directives introduced by the #if directive.
+
+if20. The chain of conditional directives ends with the #endif
+      directive.
```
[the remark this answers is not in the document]
bonachea: RESOLVED in 05-28 meeting: rename all instances to chain

bonachea:
@gklimowicz @kc9jud sectioI just rebased this PR to the tip of main to simplify git history. Please update any local clones with `git pull --rebase`

bonachea (approved):
LGTM!

Thanks for all the work on this

PR #64, opened 2025-05-07
Add a statement on whitespace needed after #ifdef, etc.

gklimowicz:
This could arguably be put somewhere else, but I thought Lorri's suggestion about Section 2.3 "Significance of whitespace" made sense. (Trying to add Lorri to reviewers but not yet succeeding.)

### DIFF
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -135,6 +135,10 @@ ws09. Whitespace characters are significant in determining token
 ws11. In fixed-form input, macro names are not recognized as such when
       spaces are inserted into their invocations.
 
+ws13. Whitespace must immediately follow the directive tokens
+      'ifdef', 'ifndef', 'elifdef', and 'elifndef' to separate them
+      from the ID tokens that follow these directive names.
+
 
 2.4 Comments
 ------------


drafts/25-xxx-specifications.txt
```diff
@@ -135,6 +135,10 @@ ws09. Whitespace characters are significant in determining token
 ws11. In fixed-form input, macro names are not recognized as such when
       spaces are inserted into their invocations.
 
+ws13. Whitespace must immediately follow the directive tokens
+      'ifdef', 'ifndef', 'elifdef', and 'elifndef' to separate them
```
kc9jud, line 139: Is this list exhaustive?

gklimowicz: I forgot about `#undef`.

Other directives can be followed by a broader set of tokens. Whitespace may not be required to separate them, and I don't have a strong opinion about whether whitespace should be required.

We already say that whitespace can be used to separate tokens, but there are places where the whitespace is not needed to separate non-identifier tokens. gfortran and flang accept the following:
```c
#define A+4
#if(1)
#endif
#include"ws003.h"
#warning"This is a warning"
```

Looking at my corpus of Fortran examples, it's not common, but it does appear (3,100 occurrences, mostly with `#if` and `#elif`).

gklimowicz, line 139:
```suggestion
      'ifdef', 'ifndef', 'elifdef', 'elifndef', and 'undef' to separate them
```

drafts/25-xxx-specifications.txt
bonachea, line 1: As previously mentioned, I dislike the entire premise of this PR. It's stating a special case  property that can be easily derived from the more general tokenization and directive recognition rules we need to state.

As I stated in [PR 61](https://github.com/j3-fortran/fpp-202y/pull/61/files#r2073880636):

> I think what we actually _mean_ here is that standard directives are only recognized as such if the token immediately following the # introducing a directive line exactly matches one of the standard directive names. If it does not match for any reason (e.g. due to lack of whitespace) then the directive does not conform to one of the standard directives and is instead parsed as a processor-dependent directive.

So for example, we should not imply the following (admittedly contrived) raw input would be _prohibited_ by the standard:
```
#ifdefpre __GNU__
#elifdefuser FOO
#elifndeftok(A,B)
#undefpre __GNU__
```
Despite their surface appearance, these are NOT syntactically prohibited forms of the `#ifdef`, `#elifdef`, `#elifndef`, `#undef` directives. After tokenization they do not match one of the specified directives, and hence they instead should be parsed as processor-dependent directive lines, with processor-dependent syntax and semantics (and hence entirely outside the scope of the standard).

bonachea (changes requested):
I disagree with this approach.

drafts/25-xxx-specifications.txt
kc9jud: Yeah, I think all of this is coming down to the fact that we haven't actually defined a token. The same kind of thing came up when we were addressing arguments to function-like macros -- do spaces around commas matter? (Do commas separate tokens? If only whitespace delineates tokens, then does that mean that a single token can be one or more argument to a function-like macro?) Right now it's unclear to me if `#if(cond)` and/or `#include<file>` should be acceptable as (standard) directives, since sometimes (maybe some??) non-alphanumeric characters seem to separate tokens implicitly.

gklimowicz: These are great points. We *do* call out the regular expression for ID in to11.

What may be missing is the link that some identifiers are treated as tokens in the illustrative syntax. That is, the `define` in a `# define` is the specific ID `define`.

Other uses of those identifier-tokens may or may not be restricted. I essentially get different results from compilers when running `-E` on this program:
```c
#define define 1
#ifdef define
#warning define defined
#endif

#define defined 1
#if defined defined
#warning defined defined
#endif
```

In any case, even if we throw out this PR (which I am OK with), we probably do need more precision in this area.

bonachea: @gak said:
> I essentially get different results from compilers when running `-E` on this program:
> 
> ```c
> #define defined 1
> ```

The line above is explicitly forbidden by C23 6.10.10-2:

> None of the following macro names in this subclause nor the identifiers `defined`, `__has_c_attribute`, `__has_include`, or `__has_embed` shall be the subject of a `#define` or a #undef` preprocessing directive.

and our own rule:

> pm12. The identifier `defined` shall not be the subject of a #define or a #undef
>      preprocessing directive.

gklimowicz:
This is not necessary, not helpful, and not going to be done.

PR #65, opened 2025-05-08
Initial draft defining syntax, semantics of controlling expressions

gklimowicz:
We describe a subset of the C constant-expression syntax for use in controlling expressions. Expression evaluation itself follows Fortran arithmetic expression semantics.

Note that the tables are a bit terse as we try to keep the line length less than the 75-character limit for J3 papers.

### DIFF
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -567,11 +567,121 @@ nd20. The result of evaluating a processor-dependent directive is processor-depe
 5 Expressions allowed in #if and #elif directives
 =================================================
 
-5.1 The 'defined' operator
---------------------------
+ex05. Controlling expressions are made up of "primary expressions" and
+      operators applied to subexpressions.
 
-6 Expression evaluation in #if and #elif directives
-===================================================
+ex10. When evaluating a #if or #elif directive, preprocessing
+      evaluates and expands all object-like macros and function-like
+      macro calls to create a token-list of the expression to be
+      evaluated.
+
+ex15. The resulting list of tokens shall be a valid expression
+      comprised of primary expressions and operators as described
+      below.
+
+ex15. Preprocessing computes the integer value of this expressions
+      using the greatest integer range available to the processor.to
+      determine the truth or falsity of the controlling expression.
+
+ex20. The evaluation of the expression shall not generate a
+      computational error (such as divide by zero).
+
+ex20. When the expression evaluates to zero, the controlling expression
+      will be considered "false". If the expression evaluates to
+      any non-zero value, the controlling expression will be considered
+      "true".
+
+
+5.1 Primary expressions
+-----------------------
+
+Preprocessing recognizes the following primary expressions in
+controlling expressions.
+
+Since expression evaluation occurs *after* token expansion, there will
+be no object-like macros or function-like macros left to evaluate. All
+instances of ID or ID (args) will all have been replaced with their
+expansions.
+
+We list them here for completeness in describing the illustrative
+syntax of controlling expressions.
+
+| Primary      | Evaluation semantics                        |
+|--------------+---------------------------------------------|
+| ID           | The expansion of the object-like macro ID   |
+| ID (args)    | The expansion of the function-like macro ID |
+| WHOLE_NUMBER | Decimal value of WHOLE_NUMBER               |
+| ( expr )     | Parenthesized expressions                   |
+|--------------+---------------------------------------------|
+
+
+
+5.1 Operators allowed in controlling expressions
+------------------------------------------------
+
+To maintain compatibility with the use of C preprocessing directives
+in many existing Fortran programs, the operators allowed in
+controlling expressions in #if and #elif expressions are a subset of
+those defined in the C 2023 standard §6.5 "Expressions" and §6.6
+"Constant expressions".
+
+A "precedence level" is assigned to each operator that determines
+how the operators combine with sub-expressions containing other operators
+at different precedence levels.
+
+An "associativity" is assigned to each operator that determines how
+operators at the same precedence level are combined.
+    - "left" means the operator binds to the left,
+    - "right" means the operator binds to the right.
+    - "nonassoc" means that the operator is not associative.
+
+The following table describes the semantics of the allowed operators
+in conditional expressions. The table is grouped by precedence level,
+from lowest precedence to highest.
+
+We label subexpressions "e1", "e2", and "e3" to aid in describing the
+evaluation semantics. Unless otherwise specified, all operators
+evaluate with the same semantics as their Fortran counterparts.
+
+
+| Prec |   Op    |    Syntax    |  Assoc'y |  Evaluation Semantics      |
+|------+---------+--------------+----------+----------------------------|
+| low  |   ? :   | e1 ? e2 : e3 |  right   |      conditional-expr      |
+|------+---------+--------------+----------+----------------------------|
+|      |   ¦¦    |   e1 || e2   |   left   |       Fortran .OR.         |
+|------+---------+--------------+----------+----------------------------|
+|      |   &&    |   e1 && e2   |   left   |       Fortran .AND.        |
+|------+---------+--------------+----------+----------------------------|
+|      |    ¦    |   e1 | e2    |   left   |    Fortran IOR(e1, e2)     |
+|      |    ^    |   e1 ^ e2    |   left   |    Fortran IAND(e1, e2)    |
+|      |    &    |   e1 & e2    |   left   |    Fortran IEOR(e1, e2)    |
+|------+---------+--------------+----------+----------------------------|
+|      |   ==    |   e1 == e2   |   left   |  1 if e1 == e2, 0 otherwise |
+|      |   !=    |   e1 != e2   |   left   | 1 if e1 /= e2, 0 otherwise |
+|------+---------+--------------+----------+----------------------------|
+|      |    >    |   e1 > e2    | nonassoc | 1 if e1 > e2, 0 otherwise  |
+|      |   >=    |   e1 >= e2   | nonassoc | 1 if e1 >= e2, 0 otherwise |
+|      |    <    |   e1 < e2    | nonassoc | 1 if e1 < e2, 0 otherwise  |
+|      |   <=    |   e1 <= e2   | nonassoc | 1 if e1 <= e2, 0 otherwise |
+|------+---------+--------------+----------+----------------------------|
+|      |   <<    |   e1 << e2   |   left   | Fortran ISHFT(e1, e2)      |
+|      |   >>    |   e1 >> e2   |   left   | Fortran ISHFT(e1, -e2)     |
+|------+---------+--------------+----------+----------------------------|
+|      |    +    |   e1 + e2    |   left   |             +              |
+|      |    -    |   e1 - e2    |   left   |             -              |
+|------+---------+--------------+----------+----------------------------|
+|      |    *    |   e1 * e2    |   left   |             *              |
+|      |    /    |   e1 / e2    |   left   |             /              |
+|      |    %    |   e1 % e2    |   left   |    Fortran MOD(e1, e2)     |
+|------+---------+--------------+----------+----------------------------|
+|      | unary + |     + e1     |  right   |        unary +             |
+|      | unary + |     - e2     |  right   |        unary -             |
+|      | defined |  defined ID  | nonassoc |    1 if the identifier     |
+|      |         |              |          |   has a #defined value,    |
+|      |         |              |          |     0 otherwise            |
+|------+---------+--------------+----------+----------------------------|
+| high | ( e1 )  |              |   N/A    |            e1              |
+|------+---------+--------------+----------+----------------------------|
 
 
 7 Predefined macros


drafts/25-xxx-specifications.txt
```diff
@@ -567,11 +567,121 @@ nd20. The result of evaluating a processor-dependent directive is processor-depe
...
+ex15. The resulting list of tokens shall be a valid expression
+      comprised of primary expressions and operators as described
+      below.
+
+ex15. Preprocessing computes the integer value of this expressions
+      using the greatest integer range available to the processor.to
+      determine the truth or falsity of the controlling expression.
+
+ex20. The evaluation of the expression shall not generate a
+      computational error (such as divide by zero).
```
kc9jud, line 587: Maybe we should be more explicit about what happens if a computational error occurs instead?
```suggestion
ex20. The processor shall reject a program if evaluation of
      the expression generates a computational error (such 
      as divide by zero).
```

drafts/25-xxx-specifications.txt
```diff
@@ -567,11 +567,121 @@ nd20. The result of evaluating a processor-dependent directive is processor-depe
...
+|------+---------+--------------+----------+----------------------------|
+|      |   ¦¦    |   e1 || e2   |   left   |       Fortran .OR.         |
+|------+---------+--------------+----------+----------------------------|
+|      |   &&    |   e1 && e2   |   left   |       Fortran .AND.        |
+|------+---------+--------------+----------+----------------------------|
+|      |    ¦    |   e1 | e2    |   left   |    Fortran IOR(e1, e2)     |
+|      |    ^    |   e1 ^ e2    |   left   |    Fortran IAND(e1, e2)    |
+|      |    &    |   e1 & e2    |   left   |    Fortran IEOR(e1, e2)    |
+|------+---------+--------------+----------+----------------------------|
+|      |   ==    |   e1 == e2   |   left   |  1 if e1 == e2, 0 otherwise |
```
kc9jud, line 659:
```suggestion
|      |   ==    |   e1 == e2   |   left   | 1 if e1 == e2, 0 otherwise |
```

drafts/25-xxx-specifications.txt
```diff
@@ -567,11 +567,121 @@ nd20. The result of evaluating a processor-dependent directive is processor-depe
...
+      evaluates and expands all object-like macros and function-like
+      macro calls to create a token-list of the expression to be
+      evaluated.
+
+ex15. The resulting list of tokens shall be a valid expression
+      comprised of primary expressions and operators as described
+      below.
+
+ex15. Preprocessing computes the integer value of this expressions
+      using the greatest integer range available to the processor.to
```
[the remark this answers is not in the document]
gklimowicz: Of course. Thank you.

drafts/25-xxx-specifications.txt
```diff
@@ -567,11 +567,121 @@ nd20. The result of evaluating a processor-dependent directive is processor-depe
...
+We list them here for completeness in describing the illustrative
+syntax of controlling expressions.
+
+| Primary      | Evaluation semantics                        |
+|--------------+---------------------------------------------|
+| ID           | The expansion of the object-like macro ID   |
+| ID (args)    | The expansion of the function-like macro ID |
+| WHOLE_NUMBER | Decimal value of WHOLE_NUMBER               |
+| ( expr )     | Parenthesized expressions                   |
+|--------------+---------------------------------------------|
```
bonachea, line 615: Minor aside: C23 6.10.2-13 perversely also allows single-character character constants as "primaries" in conditional expressions, example from C23:
```
#if ’z’ - ’a’ == 25
```
however their exact CPP evaluation semantics are implementation-defined, which means their use is not guaranteed to be portable. I don't believe I've ever seen this bizarre "feature" used in practice. 

I suspect this is some weird legacy holdover in CPP and unless someone provides a strong rationale for their inclusion I think FPP should prohibit character constants in conditional expressions.

bonachea (changes requested):
Thanks @gklimowicz for making a start on this tricky area!

Initial set of feedback:

drafts/25-xxx-specifications.txt
```diff
@@ -567,11 +567,121 @@ nd20. The result of evaluating a processor-dependent directive is processor-depe
...
+|      |    ^    |   e1 ^ e2    |   left   |    Fortran IAND(e1, e2)    |
+|      |    &    |   e1 & e2    |   left   |    Fortran IEOR(e1, e2)    |
+|------+---------+--------------+----------+----------------------------|
+|      |   ==    |   e1 == e2   |   left   | 1 if e1 == e2, 0 otherwise |
+|      |   !=    |   e1 != e2   |   left   | 1 if e1 /= e2, 0 otherwise |
+|------+---------+--------------+----------+----------------------------|
+|      |    >    |   e1 > e2    | nonassoc | 1 if e1 > e2, 0 otherwise  |
+|      |   >=    |   e1 >= e2   | nonassoc | 1 if e1 >= e2, 0 otherwise |
+|      |    <    |   e1 < e2    | nonassoc | 1 if e1 < e2, 0 otherwise  |
+|      |   <=    |   e1 <= e2   | nonassoc | 1 if e1 <= e2, 0 otherwise |
```
bonachea, line 665: In C (and hence CPP) _relational-expression_ operators are left associative.

Example:
```
#if 0 > -4 > 0
left-assoc
#else
right-assoc
#endif
```

Expands to "left-assoc" in CPP.

However it appears some FPPs may (unintentionally?) [diverge on this detail](https://godbolt.org/z/Yfzc4hnr5). Another great example why we need standardization!

```suggestion
|      |    >    |   e1 > e2    |   left   | 1 if e1 > e2, 0 otherwise  |
|      |   >=    |   e1 >= e2   |   left   | 1 if e1 >= e2, 0 otherwise |
|      |    <    |   e1 < e2    |   left  | 1 if e1 < e2, 0 otherwise  |
|      |   <=    |   e1 <= e2   |   left  | 1 if e1 <= e2, 0 otherwise |
```

drafts/25-xxx-specifications.txt
```diff
@@ -567,11 +567,121 @@ nd20. The result of evaluating a processor-dependent directive is processor-depe
...
+|------+---------+--------------+----------+----------------------------|
+| low  |   ? :   | e1 ? e2 : e3 |  right   |      conditional-expr      |
+|------+---------+--------------+----------+----------------------------|
+|      |   ¦¦    |   e1 || e2   |   left   |       Fortran .OR.         |
+|------+---------+--------------+----------+----------------------------|
+|      |   &&    |   e1 && e2   |   left   |       Fortran .AND.        |
+|------+---------+--------------+----------+----------------------------|
+|      |    ¦    |   e1 | e2    |   left   |    Fortran IOR(e1, e2)     |
+|      |    ^    |   e1 ^ e2    |   left   |    Fortran IAND(e1, e2)    |
+|      |    &    |   e1 & e2    |   left   |    Fortran IEOR(e1, e2)    |
```
bonachea, line 657: These three actually all have separate precedence in CPP

```suggestion
|      |    ¦    |   e1 | e2    |   left   |    Fortran IOR(e1, e2)     |
|------+---------+--------------+----------+----------------------------|
|      |    ^    |   e1 ^ e2    |   left   |    Fortran IAND(e1, e2)    |
|------+---------+--------------+----------+----------------------------|
|      |    &    |   e1 & e2    |   left   |    Fortran IEOR(e1, e2)    |
```

drafts/25-xxx-specifications.txt
```diff
@@ -567,11 +567,121 @@ nd20. The result of evaluating a processor-dependent directive is processor-depe
...
+|------+---------+--------------+----------+----------------------------|
+|      |    +    |   e1 + e2    |   left   |             +              |
+|      |    -    |   e1 - e2    |   left   |             -              |
+|------+---------+--------------+----------+----------------------------|
+|      |    *    |   e1 * e2    |   left   |             *              |
+|      |    /    |   e1 / e2    |   left   |             /              |
+|      |    %    |   e1 % e2    |   left   |    Fortran MOD(e1, e2)     |
+|------+---------+--------------+----------+----------------------------|
+|      | unary + |     + e1     |  right   |        unary +             |
+|      | unary + |     - e2     |  right   |        unary -             |
```
bonachea, line 678: You forgot bitwise NOT and logical NOT


```suggestion
|      | unary + |     + e1     |  right   |        unary +             |
|      | unary - |     - e1     |  right   |        unary -             |
|      | unary ~ |     ~ e1     |  right   |     Fortran NOT(e1)        |
|      | unary ! |     ! e1     |  right   | 1 if e1 == 0, 0 otherwise  |
```

drafts/25-xxx-specifications.txt
```diff
@@ -567,11 +567,121 @@ nd20. The result of evaluating a processor-dependent directive is processor-depe
...
+from lowest precedence to highest.
+
+We label subexpressions "e1", "e2", and "e3" to aid in describing the
+evaluation semantics. Unless otherwise specified, all operators
+evaluate with the same semantics as their Fortran counterparts.
+
+
+| Prec |   Op    |    Syntax    |  Assoc'y |  Evaluation Semantics      |
+|------+---------+--------------+----------+----------------------------|
+| low  |   ? :   | e1 ? e2 : e3 |  right   |      conditional-expr      |
```
bonachea, line 649: I'll note in passing that CPP also allows the comma operator in conditional expressions (at lower priority than _conditional-expression_), although it's pretty pointless in preprocessor expressions and I'm not aware of any compelling use cases. 

For this reason we (implicitly) omitted it from the requirements doc in 25-114r2. I'm only raising it now in case someone has a compelling argument to include it (something other than strict compatibility with CPP), otherwise I'm fine dropping it.

gak:
> Thanks @gak for making a start on this tricky area!

I'd like to take credit but you pinged the wrong person :)

drafts/25-xxx-specifications.txt
gklimowicz: I put them as non-associative in the Fortran preprocessor, for potential future compatibility if we ever add Fortran operators. The Fortran operators are non-associative, so we might want to disallow expressions like `0 > -4 > 0` now. (And I think chaining relational operators is terrible, but that's just my opinion. If I saw that in C code myself, I would probably replace it with something people could immediately understand.)

drafts/25-xxx-specifications.txt
gklimowicz: I left them out on purpose, but didn't have a strong reason to do so.

I may be missing a subtlety here.

In general, conditional expression evaluation is side-effect free. So, elaborating
```C
#if (my_complicated_expression, my_other_expression)
``` 
results in only the value of `my_other_expression` affecting the `#if`. `my_complicated_expression` may be evaluated, but its result is throw away.

gklimowicz: There's also the slippery slope in the C grammar, where the CPP conditional expressions start at _conditional-expression_, which I think can unfold down to _primary-expression_, which then includes the `(` _expression_ `)` which brings in the whole barnyard of comma-expressions and _assignment-expression_.

So I just chopped those rules out of the grammar.

drafts/25-xxx-specifications.txt
```diff
@@ -567,11 +567,122 @@ nd20. The result of evaluating a processor-dependent directive is processor-depe
...
+|------+---------+--------------+----------+----------------------------|
+|      |   ¦¦    |   e1 || e2   |   left   |       Fortran .OR.         |
+|------+---------+--------------+----------+----------------------------|
+|      |   &&    |   e1 && e2   |   left   |       Fortran .AND.        |
```
bonachea, line 654: In the 5/12 call we resolved these should be short-circuit evaluation as in CPP, to allow things like:
```
#if  x && 1/x
#endif
```
which means they are NOT simply  Fortran .OR. / .AND.

drafts/25-xxx-specifications.txt
```diff
@@ -567,11 +567,121 @@ nd20. The result of evaluating a processor-dependent directive is processor-depe
...
+instances of ID or ID (args) will all have been replaced with their
+expansions.
+
+We list them here for completeness in describing the illustrative
+syntax of controlling expressions.
+
+| Primary      | Evaluation semantics                        |
+|--------------+---------------------------------------------|
+| ID           | The expansion of the object-like macro ID   |
+| ID (args)    | The expansion of the function-like macro ID |
```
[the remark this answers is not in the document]
gklimowicz: Let's only describe the post-expansion grammar, and not the pre-expansion grammar. There is no such pre-expansion grammar. Dan points out that undefined ID replacement has to be done *after* processing of `##` tokens.

drafts/25-xxx-specifications.txt
```diff
@@ -567,11 +567,121 @@ nd20. The result of evaluating a processor-dependent directive is processor-depe
...
+|------+---------+--------------+----------+----------------------------|
+|      |    *    |   e1 * e2    |   left   |             *              |
+|      |    /    |   e1 / e2    |   left   |             /              |
+|      |    %    |   e1 % e2    |   left   |    Fortran MOD(e1, e2)     |
+|------+---------+--------------+----------+----------------------------|
+|      | unary + |     + e1     |  right   |        unary +             |
+|      | unary + |     - e2     |  right   |        unary -             |
+|      | defined |  defined ID  | nonassoc |    1 if the identifier     |
+|      |         |              |          |   has a #defined value,    |
+|      |         |              |          |     0 otherwise            |
```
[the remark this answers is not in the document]
bonachea: On further thought, this is just plain wrong. 
`defined` cannot appear in this table, because it needs to be applied AFTER macro expansion and BEFORE ID replacement with zero. Hence the `defined` operator (as in CPP) must be resolved and replaced before this post-expansion grammar is applied.

drafts/25-xxx-specifications.txt
kc9jud: Agreed, character constants in conditional expressions should be a (pathological) processor-dependent extension.

bonachea:
@gklimowicz @kc9jud I just rebased this PR to the tip of main to simplify git history. Please update any local clones with `git pull --rebase`

drafts/25-xxx-specifications.txt
bonachea: We resolved to exclude the comma operator, which is now documented in dfc80

kc9jud (approved):
LGTM.

bonachea (approved):
LGTM!

PR #69, opened 2025-05-19
error: Ensure program is rejected on `#error`

### LINKED ISSUE #60: Make sure that `#error` is rejected
Per discussion on 28-Apr-2025:

Right now (in #56) the behavior of `#warning` and `#error` are entirely processor-dependent. I don't think we want to leave wiggle room for implementers to simply ignore `#error` directives. Therefore a processor should be required to **reject** a program if the processor hits an `#error` directive during preprocessing (à la the language [in 4.2.2](https://j3-fortran.org/doc/year/24/24-007.pdf#section.458)).

> A Fortran processor shall reject a submitted program if the processor encounters an `#error` directive during the preprocessing phase.

kc9jud:
Fixes #60.

### DIFF
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -452,6 +452,8 @@ ew02. The token-list is not expanded.
 ew04. The processor produces a diagnostic message that includes
       the token-list.
 
+ew06. The processor shall reject a submitted program if the processor
+      encounters a `#error` directive during preprocessing.
 
 ew10. Additional behavior is processor-dependent.
 


drafts/25-xxx-specifications.txt
```diff
@@ -452,6 +452,8 @@ ew02. The token-list is not expanded.
 ew04. The processor produces a diagnostic message that includes
       the token-list.
 
+ew06. The processor shall reject a submitted program if the processor
+      encounters an `#error` directive during preprocessing.
```
gklimowicz, line 456: I believe I mostly wrote the specs as if the `#` is pronounced, so would read that last line as if it were "encounters a pound error (or hash error) directive during processing". So I used `a` instead of `an`.

```suggestion
      encounters a `#error` directive during preprocessing.
```

I notice also that we're not consistent through the doc about which we use. I filed #70 to suggest we be consistent in the final edit.

gklimowicz (changes requested):

gklimowicz (approved):

PR #72, opened 2025-05-31
Sec 4 Macro recognition and expansion: placeholder text

bonachea:
As discussed in our 2025-05-28 meeting, we want to have a placeholder in section 4 as a backup plan in case schedule prevents us from finishing a stand-alone section 4 in the next week.

This placeholder acknowledges the incomplete status, expresses our technical intent and references the relevant sections of C 2023 (which are long and subtle, but I believe the technical content really is almost exactly what we need to specify in final edits).

I've noted the few exceptions I'm aware of, where our previous discussions have indicated at least a weak consensus for FPP behavior to deliberately diverge from CPP.

### DIFF
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -577,14 +577,25 @@ me30. As specified by C 2023 6.10.5-10: "Within the sequence of
       macro, new-line is considered a normal white-space character."
 
 me31. In addition, within the sequence of preprocessing tokens making
-      up an invocation of a function-like macro, a `&` character
-      immediately followed by a newline is immediately replaced
-      by a single blank character before subsequent processing.
+      up an invocation of a function-like macro: if a `&` character is
+      immediately followed by a newline, which is immediately followed
+      by zero or more whitespace characters, which in turn are
+      immediately followed by another `&` character, then this
+      sequence of characters is immediately removed (splicing the two
+      lines and removing the `&` characters and intervening
+      whitespace) before subsequent processing.
 
-me32. The combination of rules me30 and me31 above imply that
+me32. Finally, within the sequence of preprocessing tokens making
+      up an invocation of a function-like macro: if a `&` character is
+      immediately followed by a newline (and the sequence does not
+      match the conditions of rule me31) then both are immediately
+      replaced by a single blank character before subsequent
+      processing.
+
+me33. The combination of rules me30..me32 above imply that
       when the invocation of a function-like macro is broken across
-      Fortran source lines, no line continuation character is
-      required, but that a Fortran &-style line continuation character
+      Fortran source lines, no line continuation characters are
+      required, but that Fortran &-style line continuation characters
       may optionally be used without changing the expanded result.
 
 me35. When performing macro recognition on a Fortran source line, if


bonachea:
@gklimowicz @kc9jud I spent several hours today carefully re-reading and scouring C23 6.10.5 with an eye towards any planned divergences, and this is the result. 

If this was an edit paper we really could theoretically cut-and-paste 99% of what appears in C23 6.10.5 with minor adjustments for spec terminology and these few planned behavioral adjustments, but that would result in way too much prose for this spec paper.

drafts/25-xxx-specifications.txt
```diff
@@ -555,16 +555,59 @@ nd20. The result of evaluating a processor-dependent directive is processor-depe
...
+
+me10. When determining argument boundaries in the invocation of a
+      function-like macro, FPP ignores commas surrounded by matching
+      sets of `[  ]` or `(/  /)` brackets, in addition to matching
+      sets of `(  )` parentheses (as specified in C 2023 section
+      6.10.5-11).
+
+me30. As specified by C 2023 6.10.5-10: "Within the sequence of
+      preprocessing tokens making up an invocation of a function-like
+      macro, new-line is considered a normal white-space character."
```
gklimowicz, line 577: I don't think we should have rule m30. It makes switching between function-like macros and function calls incompatible. According to this rule,
```fortran
#define atan(x,y) ATAN(x,y)
      real :: g = atan(x,
                          y)
```
and
```fortran
      real :: g = atan(x,&
                           y)
```
asymmetric. The `&` is optional in the FLM case, but required in the function case.

I think they should be required in both cases.

drafts/25-xxx-specifications.txt
```diff
@@ -555,16 +555,59 @@ nd20. The result of evaluating a processor-dependent directive is processor-depe
...
+
+me31. In addition, within the sequence of preprocessing tokens making
+      up an invocation of a function-like macro, a `&` character
+      immediately followed by a newline is immediately replaced
+      by a single blank character before subsequent processing.
```
gklimowicz, line 582: Similarly, I think me31 is problematic. Consider these two cases

```fortran
#define atan(x,y) ATAN(x,y)
      real :: g = atan(ab&
                          &c,
                          y)
```
and
```fortran
      real :: g = atan(ab&
                          &c,
                           y)
```

The Fortran continuation rules are weird; an `&` at the end of a line and at the beginning of the next line removes the whitespace between the two lines.

So me31 leaves the FLM-like macro in the first case with a dangling `&` on the next line, leaving us with the separate tokens `ATAN` `(` `ab` `&` `c` `,` `y` `)`.

The `atan` function call in the second instance in normal Fortran source becomes a call to `atan(abc, y)`. Tokens are the obvious ones.

drafts/25-xxx-specifications.txt
```diff
@@ -555,16 +555,59 @@ nd20. The result of evaluating a processor-dependent directive is processor-depe
...
+
+me32. The combination of rules me30 and me31 above imply that
+      when the invocation of a function-like macro is broken across
+      Fortran source lines, no line continuation character is
+      required, but that a Fortran &-style line continuation character
+      may optionally be used without changing the expanded result.
```
gklimowicz, line 588: If we follow my suggestions, me32 wouldn't apply.

drafts/25-xxx-specifications.txt
```diff
@@ -555,16 +555,59 @@ nd20. The result of evaluating a processor-dependent directive is processor-depe
...
+
+me35. When performing macro recognition on a Fortran source line, if
+      the sequence of preprocessing tokens making up an invocation of
+      a function-like macro includes a `!` token, the rest of the
+      current line is treated as a comment; specifically, all tokens
+      starting with that `!` up through and including the next
+      subsequent newline are immediately deleted and replaced with a
+      single blank character, before further processing of the macro
+      arguments.
```
gklimowicz, line 597: Let's talk about this one. I'm sure you've thought about this in more detail than I have tonight, but I am not sure how I get a `!` token in a replacement list that is wasn't subject to interpretation as a C `!` operator.

gklimowicz (changes requested):

kc9jud:
@bonachea this isn't going to be merged, so should we go ahead and close this PR?

PR #73, opened 2025-06-04
Reserve `#pragma STDF`

bonachea:
This is analogous to CPP 6.10.8, which reserves `#pragma STDC` for use in standardized pragmas.

### DIFF
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -632,7 +632,8 @@ pr02. The token-list may not be empty.
 3.8.2 Evaluation semantics specifications
 -----------------------------------------
 
-pr20. The token-list is expanded.
+pr20. The token-list is processed as in Fortran source fragments (see the
+      section on "Macro recognition and expansion" below).
 
 pr22. The #pragma directive causes the processor to behave
       in a processor-defined manner.
--- a/drafts/25-xxx-specifications.txt
+++ b/drafts/25-xxx-specifications.txt
@@ -929,6 +929,9 @@ dfc50. When determining argument boundaries in the invocation of a
        sets of `[  ]` or `(/  /)` brackets, in addition to matching
        sets of `(  )` parentheses.
 
+dfc60. The token-list in the FPP `#pragma` directive is always
+       subject to macro replacement, and may not be empty.
+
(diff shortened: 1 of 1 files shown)


drafts/25-xxx-specifications.txt
```diff
@@ -498,11 +498,15 @@ Example syntax:
 
 pr02. The token-list may not be empty.
 
+pr10. The token-list shall not begin with the identifier `STDF`, either
+      before or after macro expansion.
```
kc9jud, line 502: Should we also prevent the first token from beginning with `STDF`? I.e. reserve `#pragma STDF_ME` and the like?

drafts/25-xxx-specifications.txt
```diff
@@ -686,6 +690,9 @@ dfc50. When determining argument boundaries in the invocation of a
        sets of `[  ]` or `(/  /)` brackets, in addition to matching
        sets of `(  )` parentheses.
 
+dfc60. The token-list in the FPP `#pragma` directive is always
+       subject to macro replacement, and may not be empty.
```
kc9jud, line 694: Are we sure we want this? It is a divergence from CPP. If so it should be documented in the appendix.

drafts/25-xxx-specifications.txt
```diff
@@ -498,11 +498,15 @@ Example syntax:
...
+pr10. The token-list shall not begin with the identifier `STDF`, either
+      before or after macro expansion.
+
 
 3.8.2 Evaluation semantics specifications
 -----------------------------------------
 
-pr20. The token-list is expanded.
+pr20. The token-list is processed as in Fortran source fragments (see the
+      section on "Macro recognition and expansion" below).
```
bonachea, line 509:
```suggestion
pr20. The semantics of the token-list are processor-dependent. 
      In particular, it is processor-dependent whether macro 
      expansion is performed on the token-list. 
```

drafts/25-xxx-specifications.txt
bonachea:
```suggestion
dfc60. The token-list in the FPP `#pragma` directive may not be empty.
```

gklimowicz (approved):

kc9jud (approved):
LGTM.

PR #78, opened 2025-06-09
Rename draft specifications with uploaded paper number

gklimowicz:
Uploaded the draft as paper 25-142. Revisions will surely follow.


bonachea (changes requested):
I thought we had decided NOT to rename the file every time there's a new revision, to reduce git conflicts with any unmerged work.

However we SHOULD add the final revision number in the header.

Issue #62, opened 2025-05-05
To promote portability, consider adding section on preprocessing-related minimums from C 2023


gklimowicz:
C 2023 §5.2.4.1 "Translation limits" specifies the following preprocessing-related minimum values expected to be supported by C implementations.

We should probably follow suit and add something to our specifications draft. Maybe in a new section, or in section 3 on directives.

- 63 nesting levels of conditional inclusion
- 63 characters in a macro name
- 4095 macro identifiers simultaneously defined in one preprocessing translation unit
- 4095 characters in a string literal
- 127 parameters in one macro definition
- 127 arguments in one macro invocation
- 4095 characters in a logical source line
- 15 nesting levels for #included files

PR #81, opened 2025-06-12
include: Add initial text for INCLUDE line processing

### LINKED ISSUE #58: Draft specifications paper is missing a description of `INCLUDE` lines
In virtually all the discussions about `#include` preprocessing directive and the existing Fortran `INCLUDE` line, we have come to agreement that they should behave as synonyms of one another. (We fully expect implementations to provide compile-time options about this, but the standard should say they're synonymous.)

This expected behavior is missing from the draft specifications paper `25-xxx-specifications.txt`.

We need a pull request to incorporate the semantics of `INCLUDE` lines.

This was raised in the discussion of PR #56.

kc9jud:
`INCLUDE` lines should be processed very similarly to `#include` directives.

`ic10` is probably the most controversial idea here. If including a file and performing macro expansion on the included source text results in the same text, you necessarily have infinite recursion. Alternatively, resolving #62 and specifying requirements on recursion depth might be enough to handle this.

Fixes #58.


gklimowicz (approved):
LGTM

bonachea (changes requested):
Thanks for the improvements, I've added some further suggestions

kc9jud:
> **drafts/25-xxx-specifications.txt:1013**
> 
>   * Clarify how potential infinite recursion is handled when an included file expands to contain another INCLUDE line. Consider cross-referencing the discussion in #62  to provide clear guidance on recursion management.

@bonachea do we want to take Copilot's suggestion to heart?

bonachea:
@kc9jud said:
> @bonachea do we want to take Copilot's suggestion to heart?

We definitely should NOT be referencing this repo's issue tracker from a J3 paper as Copilot suggests, that seems likely to be a "bridge too far".

We already say:

> The maximum depth of nesting of any nested INCLUDE lines is processor-dependent

which I think directly implies that each compiler can deploy a reasonable finite nesting limit, and by definition any such finite limit will prevent an infinite recursion.

The limits proposed in issue #62 just provide a clearer definition of what constitutes a "reasonable" finite limit, but even without that new section I think this is still fine as-is.

bonachea (approved):
LGTM

kc9jud:
> @kc9jud said:
> 
> > @bonachea do we want to take Copilot's suggestion to heart?
> 
> We definitely should NOT be referencing this repo's issue tracker from a J3 paper as Copilot suggests, that seems likely to be a "bridge too far".
> 

Oh definitely, that's why I said "take Copilot's suggestion to heart" rather that "take Copilot's suggestion" -- it comes across as earnest if not quite getting the point...

gklimowicz (approved):
LGTM

PR #98, opened 2025-11-05
25-176: Amendments after plenary discussion of r1 and subgroup discussions

kc9jud:


drafts/25-xxx-macro-expansion.txt
```diff
@@ -29,6 +29,53 @@ specifying the details of macro identification and expansion. It should
...
+
+The subgroup does not have a recommendation and is divided. The subgroup
+unanimously agrees that either option is preferable to not passing a
+version of this paper.
+
+PROS OPTION 1:
+CONS OPTION 1:
+
+PROS OPTION 2:
+CONS OPTION 2
```
kc9jud: @bonachea any chance you can suggest some text here?

drafts/25-xxx-macro-expansion.txt
```diff
@@ -29,6 +29,53 @@ specifying the details of macro identification and expansion. It should
...
+
+OPTION 2: Fixed source form preprocessing should have recommendations
+          but be explicitly labeled as optional because of fixed-form
+          obsolescence.
+
+OPTION 3: Nothing should be specified about preprocessing for fixed
+          source form, which means preprocessing for fixed source form
+          remains processor-dependent as it is today. The edits
+          explicitly note that preprocessing of fixed source form is
+          processor dependent.
```
bonachea:
```suggestion
          remains processor-dependent as it is today. The edits should
          explicitly note that preprocessing of fixed source form is
          processor-dependent.
```

drafts/25-xxx-macro-expansion.txt
```diff
@@ -182,26 +229,83 @@ EXAMPLE age04:
...
+      required. An ampersand '&' which is the last non-whitespace
+      character (after the removal of comments) preceding a newline
+      character is removed, and all subsequent characters up to and
+      including the newline are removed. If the first non-whitespace
+      character on a continuation line is an ampersand '&', all
+      characters up to and including the leading ampersand '&' are
+      removed.
+
+lb07. [Straw poll 2, option 2] Fortran-style continuation markers are
+      permitted. An ampersand '&' which is the last non-whitespace
```
bonachea:
```suggestion
lb07. [Straw poll 2, option 2] Fortran-style continuation markers are
      permitted when the argument list spans multiple source fragment lines. 
      An ampersand '&' which is the last non-whitespace
```

drafts/25-xxx-macro-expansion.txt
```diff
@@ -182,26 +229,83 @@ EXAMPLE age04:
...
+      then the following lines:
+
+            MAC(a,
+                b,
+                c)
+            MAC(q,r
+            s,
+            t)
+
+      are erroneous.
```
bonachea: The standard basically never says anything is "erroneous". 
In particular, we don't want to _require_ an error for this case, because (despite specifying option 1) some processors may still choose to deploy this as an extension.

Strawman replacement:
```suggestion
      have a processor-dependent interpretation.
```
but there is probably a better way to state what I think we want, which is that such lines have no specified meaning but processors are not required to reject the program or generate an error diagnostic.

drafts/25-xxx-macro-expansion.txt
```diff
@@ -241,27 +345,84 @@ EXAMPLE lbe01:
 
 2.1.3 Line breaks and continuations in macro invocations (fixed-form)
 ---------------------------------------------------------------------
+[Straw poll 1, option 3 -> remove section]
```
bonachea: No I thought we decided to keep the section regardless, but mark essentially it as "rejected".

```suggestion
[Straw poll 1, option 3 -> mark subsection 2.1.3 as "rejected"]
```

drafts/25-xxx-macro-expansion.txt
```diff
@@ -381,6 +583,42 @@ EXAMPLE cme02: (free-form)
...
+             ! comment line
+           &m, n)
+            MAC(o,p ! comment 1
+           &q! comment 2
+           & r,s)
+
+      are equivalent to:
+
+      !23456
+            MAC(a,b,c)
```
bonachea:
```suggestion
```

bonachea (changes requested):
Thanks for the great start on this!

In my initial review I've made a number of suggestions, ranging from corrections for defects that I'm pretty sure are simple typos, to problems that might represent a miscommunication or disagreement on the planned amendments.

Please try to apply suggestions in the former category ASAP, and then we can continue to iterate on anything that's left.

drafts/25-xxx-macro-expansion.txt
```diff
@@ -182,26 +229,83 @@ EXAMPLE age04:
...
+      permitted. An ampersand '&' which is the last non-whitespace
+      character (after the removal of comments) preceding a newline
+      character is removed, and all subsequent characters up to and
+      including the newline are removed. If the first non-whitespace
+      character on a continuation line is an ampersand '&', all
+      characters up to and including the leading ampersand '&' are
+      removed.
+
+[Straw poll 2, option 1]
+EXAMPLE lbe01:
```
bonachea: We now have two examples labelled `lbe01`, which could be verbally confusing in discussion.
I suggest we append a unique identifier to every such duplicate example, eg:

```suggestion
EXAMPLE lbe01-A:
```

drafts/25-xxx-macro-expansion.txt
```diff
@@ -241,28 +389,85 @@ EXAMPLE lbe01:
...
       function-like macro are discarded (section 2.1.4).
 
-lc07. If character position 6 of a continuation line contains any
-      character other than a blank or zero, then any trailing blank
-      characters of the continued line are removed along with the
-      newline. Otherwise, the newline of the continued line is replaced
-      with a blank character. Characters in positions 1-6 are removed
-      from the continuation line regardless of the character in position
-      6.
+lc07. [Straw poll 2, option R] If character position 6 of a continuation
```
sblionel: Strictly speaking, a line in fixed-form that has a blank or zero in column 6 is NOT a "continuation line".  I know what you're trying to say here, but this is not right.

sblionel (changes requested):
Is there a way perhaps to adapt/reuse the existing standard text about fixed-form continuation?

drafts/25-xxx-macro-expansion.txt
```diff
@@ -29,6 +29,96 @@ specifying the details of macro identification and expansion. It should
...
+In either case, any explicit continuation characters (e.g. a trailing
+`&` in free form) are not part of the argument list of a function-like
+macro.
+
+The subgroup does not have a recommendation and is divided. The subgroup
+unanimously agrees that either option is preferable to not passing a
+version of this paper.
+
+PROS OPTION R:
+RP01. Forces a style which is more aesthetically "Fortranic".
```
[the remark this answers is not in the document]
kc9jud: @tmjbios I see that both preserve the entire string, but it seems like gfortran replaces the newline with a space.

Within a string literal, is there a way to include a literal linebreak in the literal, or do you have to use `\n`?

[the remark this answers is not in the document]
tmjbios: I disagree.

```
ag07. An argument consists of all tokens between the beginning of the
      argument list and a separating comma, between a pair of separating
      commas, or between a separating comma and the end of the argument
      list. Line continuations (sections 2.1.2 and 2.1.3), and comments
      (section 2.1.4) are removed as part of argument gathering and thus
      are not part of any arguments themselves.
```

[ag07] covers what constitutes an argument, and could possibly be updated to include newlines in addition to line continuations `&` and comments `!` as requiring special handling within an argument. Or perhaps we need to insert a new ag0X rule containing instructions on how character literals are handled as arguments?

Since [ag07] specifically states that an argument contains _all tokens_, excepting line continuation characters and comments, the newline is part of the character literal in the prior example.

Certainly we should not attempt to extend [lb05] to attempt to cover the contents of an argument - there's a distinction here we need to clear up which is the source of some ambiguity.

---

Using `\n` here so the newline is visible in this example;

```
FOO("hello",\n
"goodbye")
```

As you know, there is a difference between an individual "argument" and the "argument list". This `FOO` example is the case where a newline is within the argument list. The `LOGMSG` example posted prior does not contain a newline within the argument list - it contains a newline within the single individual argument to the FLM, and the argument is perfectly valid as it contains valid characters per ag07.

> [lb05] pretty unambiguously says that in the absence of an explicit continuation (lb07), the newline is treated as a blank character

So, no, [lb05] says no such thing as it does not cover the contents of individual arguments but is describing how to handle parsing a list when `,\n` is encountered as a possible list member.

drafts/25-xxx-macro-expansion.txt
bonachea: Resolved in 9077b9f

drafts/25-xxx-macro-expansion.txt
```diff
@@ -29,6 +29,113 @@ specifying the details of macro identification and expansion. It should
...
+Straw poll 2 analysis
+---------------------
+The analysis provided below is intended to help inform the committee
+regarding the implications of straw poll 2.
+
+PROS OPTION R:
+RP01. Forces a style which is more aesthetically "Fortranic".
+
+CONS OPTION R:
+RC01. Mandates more restrictive line continuation syntax.
```
tmjbios: We should consider adding a CON to each option, since regardless of which option is chosen a new implementation methodology is being forced on the implementer; roughly speaking it will be either "scan-then-tokenize" or "tokenize-then-replace".

Ignoring the elephant in the room seems like a bad idea, IMO.

bonachea (approved):
I think these are ready to merge and upload

kc9jud:
@bonachea this looks good to me. I will upload shortly.

